PR #10, opened 2015-04-26
make scan and utils.py pep8 compliant

esonderegger:
I'm sorry - I probably should have made this a bunch of smaller commits.

I basically kept running it through flake8 until it didn't give me any errors or warnings.

I also changed:
- the shebang to be python3
- the run command is now wrapped inside "if __name__ == '__main__':"

When you get a chance, please take a look and let me know what you think.

### DIFF
--- a/scan
+++ b/scan
@@ -241,8 +241,6 @@ def mixed(domain, other=None):
         if (os.path.exists(cache)):
             logging.debug("\tCached.")
             raw = open(cache).read()
-            data = json.loads(raw)
-
         else:
             # Only use HTTPS if it's forced. Otherwise, assume the domain
             # still needs to evaluate what mixed content appears on the
--- a/scan
+++ b/scan
@@ -102,16 +102,19 @@ def inspect(domain, other=None):
     yield [
         domain, data['canonical'], data['up'],
         data['redirect'], data['redirect_to'],
-        https_valid, data['default_https'], data['downgrade_https'], data['enforce_https'],
+        https_valid, data['default_https'], data['downgrade_https'],
+        data['enforce_https'],
         https_bad_chain, https_bad_name,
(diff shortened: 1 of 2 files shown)


konklone:
Thanks for volunteering your time to do this, @esonderegger! I have a few quibbles I noted in comments, but if those are addressed I think this is merge-able. I admit to not being a big PEP8 devotee, but I'm happy to make this project a little more approachable for people used to that style.

konklone:
One note for future work like this, is that I'd at least separate out the whitespace/indentation commit from everything else. I needed to use the `?w=1` trick on GitHub URLs to remove lines whose only changes were whitespace in order to suss out the real changes. Besides being not super widely known, that trick also disables inline commenting, so I had to jump back and forth in order to find and then note issues.

esonderegger:
Thank you for taking the time to go through my pull request so thoroughly! And thank you for suggesting pyenv. I had never used it before because the vast majority of my work has been in python 2.7. As for your concerns:

1. I agree the #!/usr/bin/env python shebang is fine because developers can choose to run the scan with:

    python3 scan whitehouse.gov

2. Inline comments are fine in pep8, they just have to have 2 spaces before them. I think the comments are nice to have and more readable inline in this case.

3. I didn't know what to do with the 'data' variable in the 'mixed' function. It seems like you may have wanted to return it at some point. I had left it commented out in my original commit because I intended to revisit it.

As for being a pep8 devotee, from a practical standpoint, I've found that running all my code through flake8 before I do anything with it has helped make it more readable and has helped me catch actual bugs. And then there's the ideological argument that, since the python language has an official style guide, I should have a good reason if I want to deviate from it. I was definitely quibbling to even suggest these changes, but I figured it was a good way to introduce myself and offer to help further on a project I think will be useful to a bunch of people.

Thanks again! And hope you've been having a great weekend!

konklone:
Thank you for taking the time to break out your commit and update the PR! Happy to merge.

PR #12, opened 2015-04-28
Refactor: restructure scanning pipeline, and separate individual scanners

konklone:
This PR does a bunch of things:

* Separates out each scanner into its own module, in a new `scanners/` directory. `utils.py` was moved into `scanners/` as well.
* Processing multiple scanners now operates _horizontally_ (runs each scan on a single domain before doing the next domain) instead of _vertically_ (runs each domain through a single scan before doing the next scan). 
* There's some aborted (added and reverted) work in this PR to add an `all.csv` that creates a single combined CSV with all data. The output mismatch between scanners (e.g. `tls` can produce 2 rows for a single domain) made this untenable.
* Data is no longer sent from `inspect` to `tls`. Instead, `tls` optionally looks to see if there's any cached data from `inspect` to help it skip known bad domains, but it can work with or without that.
* Each scanner has a couple more hooks -- an optional `init` method, an optional `command` variable, by which a scanner can initialize itself and by which the central script can verify that dependencies are installed.
* The `tls` scanner can have an alternate path to `ssllabs-scan` spelled out.
* Removes the `mixed-content` scanner, which was incomplete anyway and needs to be revisited.
* Renames `dap` to `analytics`, in preparation for proper analytics-reporter integration, and to generalize its name/function more.
* The syntax for the scanner is now `./scan whitehouse.gov --scan=inspect,tls` for both scans, and `./scan whitehouse.gov --scan=analytics --analytics=/path/to/sites.csv` for the analytics scan.
* `analytics` now outputs JSON to disk.
* Updates the docs for all of the above.

The documentation in the README is also much improved, and should keep up with all of this.


konklone:
OK, ran a complete scan under the new architecture and the data matches the last scan. Looking good. Added one more commit that fixed a bug, and merging in to begin more substantive changes to methodology.

PR #14, opened 2015-05-08
Some more pep8 changes and a small bugfix

esonderegger:
Sorry - I promised to be more involved and then dropped off the face of the earth for almost two weeks.

I have a bunch of small style guide commits for you, separated out by type. They should be fairly self-explanatory.

The last one is a mis-named parameter to logging.error that only gets triggered if the user doesn't have the command on their system. I ran into it when I tried to a tls scan before installing the ssllabs-scan tool.

Thanks again for building this tool! I'm working on compiling a csv file of .mil domains to run against it to see just how poorly the military is doing at https.

### DIFF
--- a/scan
+++ b/scan
@@ -55,6 +55,7 @@ def run(options=None):
 
     scan_domains(scans, domains)
 
+
 ###
 # Given the selected scanners, and input domains, run each domain
 # through each scanner.
@@ -93,6 +94,7 @@ def scan_domains(scanners, domains):
         handles[scanner]['file'].close()
 
     logging.warn("Results written to CSV.")
+
 
 # Yield domain names from a single string, or a CSV of them.
 def domains_from(arg):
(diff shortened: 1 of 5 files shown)


konklone:
> I'm working on compiling a csv file of .mil domains to run against it to see just how poorly the military is doing at https.

Please share when you're done! I spent a bit of time on that myself a few months back -- here's [a gist of the 65 I dug up](https://gist.github.com/konklone/f8e28b5a0b1f15110e0c). I'm sure it's *extremely* incomplete, and doesn't even attempt to track subdomains.

konklone:
Thanks for doing this! And I hate to have made more work for you -- if you want to do a PR that makes Travis builds depend on passing a PEP8 check, I'd merge that too.

I left one negative comment venting my frustration about line length limits (the world is a beautiful and ever-evolving place that shouldn't be held back by 30-year old conventions of terminal-based text editors), but I'm not going to make it a dealbreaker.

PR #15, opened 2015-05-15
Dockerize

vzvenyach:
This PR gives a user the option of running the `domain-scan` scanners as a service using Docker. Additionally, to ease the use of Docker, the PR uses `docker-compose` to allow a user to simply `docker-compose up` and then `docker-compose run scan ...`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,54 @@
+# VERSION 0.0.1
+
+# USAGE
+
+FROM      ubuntu:14.04
+MAINTAINER V. David Zvenyach <vladlen.zvenyach@gsa.gov>
+
+COPY . /tmp/ 
+WORKDIR /tmp
+
+#RUN apt-get update && apt-get install -y python3-pip
+#RUN pip3 install -r requirements.txt
+
+# Install ruby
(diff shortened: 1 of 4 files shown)


scan
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
konklone: Is this required?

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# Install Bundler for each version of ruby
+RUN /bin/bash -l -c "gem install bundler"
+RUN /bin/bash -l -c "gem install specific_install"
+RUN /bin/bash -l -c "gem specific_install -l http://github.com/benbalter/site-inspector.git -b erics-mode"
+
+# Install Go
+RUN apt-get update && apt-get install -y \
+		gcc libc6-dev make \
+		--no-install-recommends \
+	&& rm -rf /var/lib/apt/lists/*
```
konklone: Curious, do you know why the `rm -rf /va/lib/apt/lists/*`?

konklone:
Tested and this _totally_ works. Thanks, @vzvenyach! Also, I've given you write access, so you can do future PRs on branches.

Dockerfile
ozzyjohnson: It's about saving space in the image. It's good for 10s to 100+ MB, a huge fraction when you're looking at a <1GB image.

Issue #16, opened 2015-05-29
Parallelize domain scanning


konklone:
Doing this all in serial is a big waste, especially since the requests are hitting different domains. Parallelizing the tasks won't increase load on any scanned site, but will drastically speed up completion time.

https://docs.python.org/3/library/asyncio-dev.html

PR #21, opened 2015-06-09
Added ga-scanner

ramirezg:
Added the google analytics scanner, but I couldn't find a good way to install PhantomJS though the Dockerfile. Any thoughts?


scan
```diff
@@ -80,13 +80,15 @@ def scan_domains(scanners, domains):
             'writer': scanner_writer
         }
 
-    # For each domain, run through each scanner.
```
ramirezg: The loop structure was flipped so I could keep the same instance of phantomjs open while scanning.

konklone: Flipping the loop structure is a major change that I'd prefer to tackle separately. 

Additionally, this project is moving towards parallelizing the scanning process (#16), so keeping open a single instance of phantomjs is not desirable. I'd rather accept an annoyingly slow Google Analytics scanner, and use that as an impetus to complete the parallelization process, which will make the whole process a *lot* faster for both this new scanner and the existing scanners.

scan
```diff
@@ -80,13 +80,15 @@ def scan_domains(scanners, domains):
...
-            # A scanner can return multiple rows.
+    # A scanner can return multiple rows.
+    for scanner in scanners:
+        # For each domain, run through each scanner.
+        for domain in domains_from(domains):
             for row in scanner.scan(domain, options):
                 if row:
                     handles[scanner]['writer'].writerow([domain] + row)
+        if scanner.__name__ == 'scanners.googleanalytics':
+            scanner.clean_up()
```
konklone: Instead of hardcoding in an exception here, the scanning process should look to see if a scanner defines a `cleanup()` method, and if it does, execute it.

However, if we follow my above note and restore the scanning loop, this scanner won't need to hold open a single browser instance anymore.

scanners/googleanalytics.py
```diff
@@ -0,0 +1,115 @@
...
+    console.log('ga_version: No Google Analytics');
+    console.log('ga_ua_code: No Google Analytics');
+    console.log('ga_anon_ip: No Google Analytics');
+    console.log('ga_force_ssl: No Google Analytics');
+}
+        """)
+
+    def clean_message(self, element, message):
+        """ Cleans the message log """
+        return message.split(element)[-1].replace(' (:)', '').strip(': ')
```
konklone: Instead of parsing the message using colons, the JavaScript should make a JSON object and use a single `console.log` command to spit it out. This will make parsing safer, more reliable, and will map directly to a Python dict.

scanners/googleanalytics.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def __init__(self):
+        d = DesiredCapabilities.PHANTOMJS
+        d['loggingPrefs'] = {'browser': 'ALL'}
+        self.browser = webdriver.PhantomJS(desired_capabilities=d)
+
+    def get_url(self, url):
+        """ Get URL fron browser """
+        self.browser.get(url)
+        time.sleep(3)
```
konklone: Why does this need to `sleep()` here? And is this 3 seconds, or 3 milliseconds?

scan
konklone: A note that I just tackled parallelization in #22, which did involve flipping the loop after all. However, I'm not sure of the effect this parallelization will have on phantomjs. Happy to work with you to figure it out.

scanners/googleanalytics.py
ramirezg: Changed.

ramirezg:
@konklone I think this update should make the ga scanner work with PR #22 by creating and cleaning up a `GAChecker` instance on each scan. 

Next I plan to look at the issues below

* When driving PhantomJS, I tend to use the built-in JavaScript interfaces. (In past projects I've run into frustrating Selenium-related issues, so I try to avoid the wrapper.) I don't think it matters too much.

* Google Analytics might not get instantiated as window.ga (or window._sendPageView). The ga approach is definitely the norm, but there are plenty of other implementations in the wild. For a more robust check, you could also instrument HTTP(S) requests to the Google Analytics reporting endpoints.

garrettr:
> I couldn't find a good way to install PhantomJS though the Dockerfile. Any thoughts?

FWIW, my PR for for #31 adds PhantomJS to the Dockerfile.

konklone:
@ramirezg Sorry for letting this sit, but do you mind updating me on your latest thinking? Should I evaluate the current PR for inclusion?

ramirezg:
@konklone So the the ga-scanner has been working for 18f sites. However, it seems that outside 18F, GA might not get instantiated as ` window.ga` so I think this might be more useful as an internal tool.

konklone:
OK, I'm going to close this for now, but feel free to re-open if you want to take this further.

PR #23, opened 2015-06-14
open domain.csv as utf-8

coolya:
I had some problems opening domain.csv files that where stored as UTF-8. Not sure if it was a problem of the docker container or a general thing. I added the encoding explicitly. If this breaks any existing domain.csv I can a some detection logic for the encoding to avoid this.

### DIFF
--- a/scan
+++ b/scan
@@ -112,7 +112,7 @@ def scan_domains(scanners, domains):
 # Yield domain names from a single string, or a CSV of them.
 def domains_from(arg):
     if arg.endswith(".csv"):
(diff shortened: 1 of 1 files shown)


konklone:
Seems all-upside to me!

konklone:
@coolya Do you see any issues with using `encoding='utf-8'` when writing CSVs? Seems like we should probably do it on both sides.

coolya:
I don't see any issue with that. 
The CSV that was casing the problems was written with sublime text. But writing them in UTF will probably avoid problems with our german names which have umlauts.

PR #24, opened 2015-06-14
Change Dockerfile to use site-inspector 1.0.2

robbi5:
The readme said "This repo depends on site-inspector 1.0.2" but the Dockerfile tries to install an no longer existing (?) branch of site-inspector.


parkr:
:+1:

konklone:
Nice catch! I didn't realize the branch was gone, but makes sense now that https://github.com/benbalter/site-inspector/issues/38 is taken care of. Thanks!

PR #31, opened 2015-06-22
Pageload scanner

garrettr:
This PR adds a scanner for page load and rendering times based on [Phantomas](https://github.com/macbre/phantomas). Currently it only returns an (arbitrarily selected) subset of metrics related to page load performance, but there are [many useful metrics available](https://www.npmjs.com/package/phantomas#metrics).

I got the idea for this from the [Pulse README](https://github.com/18F/pulse/blob/master/README.md), where "Ideas for later versions" includes "load time (server-side)". It looks like it should also be easy to implement some of the other ideas on that list with Phantomas, including:

* Mixed content detection
* Use of third party services


scanners/pageload.py
```diff
@@ -0,0 +1,42 @@
...
+
+command = Phantomas.CMD
+init = None
+
+def scan(domain, options):
+    logging.debug("[%s][pageload]" % domain)
+
+    if not (domain.startswith('http://') or
+            domain.startswith('https://')):
+        domain = 'http://' + domain
```
konklone: Oddly enough (at least in the heady world of `.gov` domains) some domains don't bother setting a DNS entry for their root domain, or their redirect doesn't work correctly. One thing the `inspect` task does is determine the "canonical" endpoint for a domain, so we could have it look for the presence of a `cache/inspect/[domain].json` file and, if present, use it to determine what URL to construct.

We're doing something similar [in the tls scanner](https://github.com/18F/domain-scan/blob/master/scanners/tls.py#L24-L28) to not bother running SSL Labs scans on domains which we know don't have HTTPS.

konklone:
In any case, none of my comments on these two PRs are merge-blockers -- I'll do a full test and merge shortly.

scanners/pageload.py
garrettr: Yeah, this was a bit of a hack because Phantomas needs the full URL (w/ scheme) to work properly, and I wanted to maintain compatibility with the other scanners (e.g. `scan --scanners=tls,pageload whitehouse.gov` should work). If the `inspect` task is already determining the endpoint in a more intelligent way, then I think it'd be great to re-use those results here.

My only concern is that it would create a dependency between two of the scanners, and my thinking was that they should all be independent. However, we could easily use the `inspect` data if it is available (as you do in the linked TLS scanner code), and otherwise fall back onto this existing logic.

konklone: > However, we could easily use the inspect data if it is available (as you do in the linked TLS scanner code), and otherwise fall back onto this existing logic.

Exactly.

konklone:
I'm working on this now -- I'm going to cut out the Python dependency, and ask people to always install the npm version. I wasn't able to get the `phantomas` binary to be automatically added to the path through a `pip install`, and I'd prefer to simplify the code path inside the tool. I also generally like shell-ing out to commands where possible, to keep the actual code we maintain in domain-scan primarily about coordination of external tools.

I'm also finding that phantomas doesn't properly handle websites that require SNI support, because it depends on phantomjs 1.X. phantomjs 2 is supposed to do this better, but updating phantomas to work with it is a work in progress. So, I'm going to update the `pageload` scanner to pass a flag that ignores TLS errors (which is probably helpful anyway, so that it evaluates sites which have an invalid certificate chain but which browsers generally load correctly anyway -- there's a significant contingent of websites that are in this situation).

konklone:
I've merged this PR in, along with some further work that centers on the CLI interface to phantomas, ignores SSL errors, integrates the project's standard caching mechanism, and uses the canonical endpoint where possible:

* https://github.com/18F/domain-scan/commit/14ffe54273f2166d987a692c0a1c8e0a6d9568fc
* https://github.com/18F/domain-scan/commit/e417549cbe7e6455fbfa3819ac3161b771d5c3ed
* https://github.com/18F/domain-scan/commit/ef793ef251577544ff12d14034f8026eb1ae93a5
* https://github.com/18F/domain-scan/commit/c2ec9367f8114407f8a42ed27ec640abb7db8532
* https://github.com/18F/domain-scan/commit/46ea6b3c64d3d0ec4a2dd69e7eee1a4b82da8655
* https://github.com/18F/domain-scan/commit/4b9d80dd9d321a55b3084ae22324fd5384f4d949
* https://github.com/18F/domain-scan/commit/2dab9ec9e7defeeb9aab2bc0574e5f536b69a30e
* https://github.com/18F/domain-scan/commit/669025df0333751a45b4937796a41f07a4569acd

(I'll do a separate PR next time if I do so many follow-up commits.)

I'm running `pageload` on the `.gov` domain list now, which might flush out some more error conditions. Thanks for the submission, @garrettr!

konklone:
I also just added 888b1bc, which skips domains that `inspect` detected as either down, or which just as redirects to external domains.

garrettr:
> I'm also finding that phantomas doesn't properly handle websites that require SNI support, because it depends on phantomjs 1.X. phantomjs 2 is supposed to do this better, but updating phantomas to work with it is a work in progress. So, I'm going to update the pageload scanner to pass a flag that ignores TLS errors (which is probably helpful anyway, so that it evaluates sites which have an invalid certificate chain but which browsers generally load correctly anyway -- there's a significant contingent of websites that are in this situation).

I'd just noticed that myself and was about to comment here :) Your workaround sgtm.

garrettr:
In addition, I just pushed a commit making the change [we discussed](https://github.com/18F/domain-scan/pull/31#discussion_r33506538) to use the canonical hostname from the inspect data if available (and otherwise fall back on prepending `http://` if `domain` isn't a full URL).

You merged this PR in the meantime, but if you're interested in including that change as well, it's here: https://github.com/garrettr/domain-scan/commit/af0fa4b5c69df8fc499772aeba5cc5edb2ebdfd5 (or I can make a new PR for it, if you want).

Will probably need manual merging given all the changes you made, but should be pretty easy because the changes are small and localized.

konklone:
@garrettr Do you have any idea why this is timing out?

```
phantomas https://www.usajobs.gov --reporter=json
```

Adding `--verbose` shows more output, but I can't really tell why it keeps cycling through elements and never quits before the timeout. I've tried upping the timeout to 30s (it defaults to 15s) but I get the same thing.

Also, I defaulted the worker pool for `pageload` to 2, since having 10 of these running in parallel by default would, I suspect, add enough local resource contention to start having an effect on the measurements.

konklone:
> In addition, I just pushed a commit making the change we discussed to use the canonical hostname from the inspect data if available (and otherwise fall back on prepending http:// if domain isn't a full URL).

That was included in my work too, so it should be all set.

garrettr:
> @garrettr Do you have any idea why this is timing out?

No idea. Have you tried the `--debug` flag?

garrettr:
FWIW, I think there is a problem with the "don't cache invalid results" logic on `scanners/pageload.py:61`. When I tried duplicating your error with `https://usajobs.gov` on my end, the result was cached even though there was an error running phantomas.

```sh
➜  domain-scan git:(master) docker-compose run scan https://www.usajobs.gov --scan=pageload
phantomas: (252) Timeout
Error running ['phantomas', 'https://www.usajobs.gov', '--reporter=json', '--ignore-ssl-errors'].
Results written to CSV.
➜  domain-scan git:(master) docker-compose run scan https://www.usajobs.gov --scan=pageload
Results written to CSV.
➜  domain-scan git:(master) rm cache/pageload/https:/www.usajobs.gov.json
➜  domain-scan git:(master) docker-compose run scan https://www.usajobs.gov --scan=pageload
phantomas: (252) Timeout
Error running ['phantomas', 'https://www.usajobs.gov', '--reporter=json', '--ignore-ssl-errors'].
Results written to CSV.
```

garrettr:
@konklone I ran it with `--verbose` (puts Phantomas in debug mode) and  `--debug` (puts PhantomsJS in debug mode) and got a bunch of log, but the cause of the problem didn't immediately pop out at me. I'd say you should file an issue upstream with Phantomas.

konklone:
@garrettr The behavior of caching "invalid" results is, for better or worse, intended. I'm not a huge fan of some of the results of this behavior, but caching is meant to allow resuming of large batch scans from where they left off. Suggestions for modifying this behavior are very welcome.

PR #32, opened 2015-06-23
Check if domain is in the Chrome HSTS Preload list

### LINKED ISSUE #30: Measure whether a domain is actually in the Chrome preload list
Get the Chrome preload list from source control somehow, and measure whether the domain is actually in there. This can help eliminate the gap between what we think is preload-ready and what actually made it.

garrettr:
This is a first pass at implementing this feature. Resolves #30.

Test: `docker-compose run scan torproject.org --scanner=tls`. There is a new "In Chrome Preload List" header in `results/tls.csv` that should be `True` for `torproject.org`.

Some additional things that should be addressed, but I would appreciate feedback on, are:

1. This PR fetches the Chrome preload list every time `scanners.tls.scan` is run. Since the preload list does not change that frequently, this is inefficient and incurs an `O(n)` penalty for `n` lookups. In practice, fetching the list is quite fast, and the lookup time is dominated by the far slower SSLLabs API request.

    The best solution is probably to fetch the Chrome preload list at the beginning of the TLS scan run, and use the same list for all domains that are checked during that run. The only reason I haven't implemented this is a bit of uncertainty about the right way to do it. It would be easy to memoize the lookup so it only happens on the first TLS scan, but the only available place to store that right now is in the global scope on the `scanners.tls` module and this is generally a bad idea.

    Re-factoring the `scanners.tls` module into a `TLSScanner` class is one approach that comes to mind. We could fetch the Chrome preload list once when instantiating the class in `__init__`, then re-use that instance for as many lookups as we need.
2. This PR fetches the preload list from Chrome's master branch. I have not looked into the details of their release schedule, but presume that this list could be as much as 6 weeks ahead of what the vast majority of Chrome users are using in their release-channel browsers.

    Whether we care about fetching the list from the latest release branch or the master branch is a question whether we want to know "this website had HSTS preload for the vast majority of Chrome users using the standard release branch when we checked", or "this site has been accepted into the Chrome preload list, and will eventually be HSTS preloaded for all Chrome users".

### DIFF
--- a/scanners/tls.py
+++ b/scanners/tls.py
@@ -2,6 +2,9 @@ import logging
 from scanners import utils
 import json
 import os
+import base64
+import urllib.request
+import re
 
 
 ###
@@ -17,6 +20,27 @@ import os
 command = os.environ.get("SSLLABS_PATH", "ssllabs-scan")
 init = None
 workers = 5
+
+def get_chrome_preload_list():
+    preload_list_url = ('https://chromium.googlesource.com/chromium/src/net/+'
+                        '/master/http/transport_security_state_static.json')
+    with urllib.request.urlopen(preload_list_url + '?format=text') as response:
+        raw = response.read()
+
+    # To avoid parsing the contents of the file out of the source tree viewer's
+    # HTML, we download it as a raw file. googlesource.com Base64-encodes the
+    # file to avoid potential content injection issues, so we need to decode it
+    # before using it. https://code.google.com/p/gitiles/issues/detail?id=7
+    raw = base64.b64decode(raw).decode('utf-8')
+
+    # The .json file contains '//' comments, which are not actually valid JSON,
+    # and confuse Python's JSON decoder. Begone, foul comments!
+    raw = ''.join([ re.sub(r'//.*$', '', line)
+                    for line in raw.splitlines() ])
+
+    preload_list_json = json.loads(raw)
+    return { entry['name'] for entry in preload_list_json['entries'] }
+
 
 def scan(domain, options):
     logging.debug("[%s][tls]" % domain)
@@ -70,6 +94,10 @@ def scan(domain, options):
                 return None
                 # raise Exception("Invalid data from ssllabs-scan: %s" % raw)
 
+        # TODO: The Chrome preload list changes relatively infrequently, so we
+        # should probably cache it. Loading it for every `scan` is inefficient.
+        chrome_preload_list = get_chrome_preload_list()
+
         # can return multiple rows, one for each 'endpoint'
         for endpoint in data['endpoints']:
 
@@ -107,7 +135,8 @@ def scan(domain, options):
                 tlsv12,
                 spdy,
                 endpoint['details']['sniRequired'],
-                h2
+                h2,
+                domain in chrome_preload_list,
             ]
 
 headers = [
@@ -117,5 +146,6 @@ headers = [
     "Fallback SCSV",  # good things
     "RC4", "SSLv3",  # old things
     "TLSv1.2", "SPDY", "Requires SNI",  # forward
-    "HTTP/2"  # ever forward
+    "HTTP/2",  # ever forward
+    "In Chrome Preload List"
 ]
--- a/scanners/tls.py
+++ b/scanners/tls.py
@@ -16,12 +16,15 @@ import re
 # and only process domains with valid HTTPS, or broken chains.
 ###
 
-
 command = os.environ.get("SSLLABS_PATH", "ssllabs-scan")
-init = None
 workers = 5
 
+chrome_preload_list = None
+
+
 def get_chrome_preload_list():
+    logging.debug("Fetching chrome preload list...")
+
     preload_list_url = ('https://chromium.googlesource.com/chromium/src/net/+'
                         '/master/http/transport_security_state_static.json')
     with urllib.request.urlopen(preload_list_url + '?format=text') as response:
@@ -40,6 +43,17 @@ def get_chrome_preload_list():
 
     preload_list_json = json.loads(raw)
     return { entry['name'] for entry in preload_list_json['entries'] }
+
+
+def init(options):
+    """
+    Download the Chrome preload list at the beginning of the scan, and
+    re-use it for each scan. It is unnecessary to re-download the list for each
+    scan because it changes infrequently.
+    """
+    global chrome_preload_list
+    chrome_preload_list = get_chrome_preload_list()
+    return True
 
 
 def scan(domain, options):
@@ -93,10 +107,6 @@ def scan(domain, options):
             else:
                 return None
                 # raise Exception("Invalid data from ssllabs-scan: %s" % raw)
-
-        # TODO: The Chrome preload list changes relatively infrequently, so we
-        # should probably cache it. Loading it for every `scan` is inefficient.
-        chrome_preload_list = get_chrome_preload_list()
 
         # can return multiple rows, one for each 'endpoint'
         for endpoint in data['endpoints']:


konklone:
This is awesome, @garrettr, sorry it's taken me so long to triage this!

> The best solution is probably to fetch the Chrome preload list at the beginning of the TLS scan run, and use the same list for all domains that are checked during that run. The only reason I haven't implemented this is a bit of uncertainty about the right way to do it. It would be easy to memoize the lookup so it only happens on the first TLS scan, but the only available place to store that right now is in the global scope on the `scanners.tls` module and this is generally a bad idea.

I personally don't mind memoizing the lookup on the global scope. `*shrug*` But another way to do it would be to check for it on disk in a predictable place in the current report's `cache/` directory, and only download it if it's not present. The `--force` flag is already used to avoid caching (and to send a flag to SSL Labs to invalidate the cache) and could be used to force a re-download no matter what.

> Whether we care about fetching the list from the latest release branch or the master branch is a question whether we want to know "this website had HSTS preload for the vast majority of Chrome users using the standard release branch when we checked", or "this site has been accepted into the Chrome preload list, and will eventually be HSTS preloaded for all Chrome users".

For my case at least, it's definitely the latter, since preparation and acceptance takes the most work, and is kind of all we can be certain about. For example, Mozilla has their own criteria for deciding which domains in the Chrome preload list should make it into Firefox, which involves re-scanning all the domains for existing HSTS headers. That could change at any time, and I'm not sure what Safari's or IE's processes are or will be, or if as closed-source browsers they'll even document them.

> This PR fetches the Chrome preload list every time scanners.tls.scan is run. Since the preload list does not change that frequently, this is inefficient and incurs an O(n) penalty for n lookups. In practice, fetching the list is quite fast, and the lookup time is dominated by the far slower SSLLabs API request.

This could be done as its own stand-alone scanner, since we don't need the results of the SSL Labs API as a precondition to examining the preload list. I'm not too concerned about it either way for now.

garrettr:
Cool! Is this ready to merge, or do you want me to resolve the caching issue first? I'm happy to take a stab at either memoizing it in the module level and/or caching it to disk.

konklone:
It's not a requirement, but if you want to give caching a go first, that'd be great.

garrettr:
@konklone It seems like the optional `init` function would be the right place to download and cache the preload list.

garrettr:
I'm working on adding the caching in `init` now, so don't merge this just yet!

konklone:
:+1: I won't be testing for merge until at least tomorrow, so just let me know when I should take another look.

garrettr:
@konklone Done! Now it just fetches the preload list once every time `scan` is called, instead of once for each domain that gets scanned. We could extend this further by caching the preload list to disk and only refreshing it if it gets stale (according to some metric) but I think this is good enough for now.

scanners/tls.py
```diff
@@ -13,11 +16,46 @@
...
+
+def init(options):
+    """
+    Download the Chrome preload list at the beginning of the scan, and
+    re-use it for each scan. It is unnecessary to re-download the list for each
+    scan because it changes infrequently.
+    """
+    global chrome_preload_list
+    chrome_preload_list = get_chrome_preload_list()
+    return True
```
garrettr: This was a bit weird - I had to return a "truthy" value from the scanner's init function, or else `scan` would exit silently without any error or info message. This is due to the following code from `scan`:

```py
        # Scanners can have an optional init/validation hook.
        if scanner.init and (not scanner.init(options)):
            exit()
```

I'm fine with this, but it might be nice to document or explain what the intention is here. Would it be better to use exceptions instead of return values for communicating error conditions here?

garrettr:
@konklone Does this need anything else in order to be merged?

konklone:
@garrettr OK, finally merged. This was super helpful, and I will be a lot more responsive going forward.

I made a few additional commits for the merge:

* https://github.com/18F/domain-scan/commit/6eaa0f6af21d75969a52fddce7a0f000d4033b5e - aesthetics
* https://github.com/18F/domain-scan/commit/ae193bef7d578097c1940b2219f5c0bb426adc8c - note when validation fails (not really applicable here, but I just noticed it)
* https://github.com/18F/domain-scan/commit/16391473241af40f085373f018f3e81b57060ff1 - moved the process from `tls` to `inspect`.

I moved the preload determination from `tls` to `inspect` for three main reasons:

* The other HSTS information is already calculated and returned during the `inspect` process.
* The `tls` scanner outputs a row for each individual endpoint, and HSTS information is per-domain.
* It's very convenient to not have to wait for an SSL Labs scan to complete in order to get this data. The `inspect` process is orders of magnitude quicker.

Let me know if you have issues with the changes. I'm planning to integrate this into [pulse.cio.gov](https://pulse.cio.gov) so that it indicates whether or not a domain is present in the list.

konklone:
I also just fixed a crash in https://github.com/18F/domain-scan/commit/0e518f4af375a3611cae1713aeaab67b11d44583 that occurred after I removed the `init` attribute from `tls` in the transition process.

konklone:
@garrettr So here's a question -- should we say that a domain is "in the preload" list in the column we generate, if the domain's policy in the preload list does not include subdomains? This includes `healthcare.gov`, for instance.

The domain's root endpoint is in the preload list, but conceptually, I'd argue the **domain** is not in the preload list unless its subdomains are considered included as well.

If we agree on this, then the scanner should be changed to look at this sub-policy and mark the column with a `False` result. Alternatively, we could change it from a boolean field to a string field with 3 possible values, so as to capture the "partial" result. I'm more inclined to just say `False`.

garrettr:
@konklone (re `include_subdomains`) Yeah, the entry for `healthcare.gov` is a weird one. It doesn't have any `include_subdomains` key at all (I assume it defaults to false), while there are other entries that explicitly set `'include_subdomains': false`. The discrepancy is odd -- I wonder if it's worth filing a bug on Chromium?

I agree that a domain should not be considered preloaded unless all of its subdomains are as well. Not including subdomains can also lead to attacks on cookies and an interesting [privacy leak](http://www.leviathansecurity.com/blog/the-double-edged-sword-of-hsts-persistence-and-privacy).

I am fine with either resolution. I think it might be nice to communicate to admins that there's a "partial" stage, since that rewards them for setting the header in the first place and might prompt them to work on getting their site to work with `includeSubdomains`. But I'd defer to you in either case.

garrettr:
The output from `inspect` already has a `HSTS All Subdomains` field. It sounds like we could use that to capture the partial result, e.g. if a site sets HSTS without `includeSubdomains` you could say `HSTS All Subdomains == False` but leave `HSTS` and `HSTS Preloaded` `== True`.

Of course, that would imply you would need some logic in your view to figure out how to convey this, which you might not want.

konklone:
> Yeah, the entry for healthcare.gov is a weird one. It doesn't have any include_subdomains key at all (I assume it defaults to false), while there are other entries that explicitly set 'include_subdomains': false. The discrepancy is odd -- I wonder if it's worth filing a bug on Chromium?

That is odd. I'm certain the default is `false`, but the difference in implicitness is weird. I'll cc @lgarron if he thinks tt's worth cleaning up.

> I am fine with either resolution. I think it might be nice to communicate to admins that there's a "partial" stage, since that rewards them for setting the header in the first place and might prompt them to work on getting their site to work with includeSubdomains. But I'd defer to you in either case.

At least on pulse.cio.gov, we do give "partial credit" for having HSTS set for the base domain, even if subdomains aren't included. I feel like for preloading, there's less need for partial credit, and more to gain by holding out from saying it's preloaded until it's _fully_ preloaded.

> The output from inspect already has a HSTS All Subdomains field. It sounds like we could use that to capture the partial result, e.g. if a site sets HSTS without includeSubdomains you could say HSTS All Subdomains == False but leave HSTS and HSTS Preloaded == True.

That `HSTS All Subdomains` field is a measurement of the HSTS header, not the preload line, and it's not safe to assume that one follows from the other (or that a team even *wants* to be preloaded, even if they've got a HSTS header with `includeSubDomains`).

I think it's interesting from a _data_ perspective to have some level of granularity to indicate "partially preloaded", but since I don't want to put it in the UX at all, I'm not very motivated to add the complexity there, and lean towards just doing `true`/`false`.

Analogously, I spent some serious time early in development distinguishing between "enforces HTTPS" (eventually redirects to HTTPS) and "strictly enforces HTTPS" (_immediately_ redirects to HTTPS). But in practice, I found that what I cared about motivating most was just the former -- I have no desire to punish anyone who didn't make it to the super-strict form -- and so I have never used the data I distinguished by measuring the latter in any way. I probably just shouldn't have bothered spending all the time on capturing the data that I did.

PR #33, opened 2015-07-01
Add STARTTLS scanner based on starttls.info

garrettr:
Supports the eventual resolution of https://github.com/18F/pulse/issues/218.

### DIFF
--- a/scanners/starttls.py
+++ b/scanners/starttls.py
@@ -1,0 +1,150 @@
+from datetime import datetime, timedelta
+import logging
+import re
+import requests
+import time
+
+command = None
+init = None
+
+
+def starttls_check_url(domain):
+    return 'https://starttls.info/api/check/%s' % domain
+
+
+def poll_starttls_info_status(domain):
+    """
+    Poll the starttls.info API endpoint until the status of the info for a
+    given `domain` is 'DONE'.
+
+    We poll every 5 seconds.
+
+    # TODO: set a maximum poll time after which we stop trying and report an error
+    """
+    poll_interval = 5
+    while True:
+        r = requests.get(starttls_check_url(domain))
+        if r.json()['status'] == 'DONE':
+            return r
+        time.sleep(poll_interval)
+
+
+def refresh_starttls_results(domain):
+    """
+    Send a `reset` POST request to the STARTTLS API endpoint to ask the
+    server to refresh the results for `domain`.
+
+    After the reset request is sent, we poll the endpoint until it is done
+    refreshing the results, then return the complete set of new results.
+    """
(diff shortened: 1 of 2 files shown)


garrettr:
This scanner is not perfect, but that is mostly due to issues in `starttls.info`. We are [actively encouraging the original developer to open-source their code](https://github.com/18F/pulse/issues/218#issuecomment-112324244), and are hoping they will do so soon so we can contribute and help them improve it.

Our thinking is that it makes sense for a separate site/project (not Pulse, not domain-scan) to be the canonical place to go for STARTTLS scanning and grading, and starttls.info has made a great start on that front. We'd like to encourage and improve that project independently, since it's useful generally and not just for the Pulse/domain-scan use case. Ideally, we would be able to link out to the `starttls.info` score page for a given domain in the same manner that we currently do for the SSL Labs score.

konklone:
This is great, and I don't have a problem with pinging starttls.info, if @einaros doesn't have terms of service limitations on the API (this is separate from open/closed source). For example, SSL Labs isn't just closed source, but they also have [Terms](https://github.com/ssllabs/ssllabs-scan/blob/stable/ssllabs-api-docs.md#terms-and-conditions) that prohibit the public disclosure of grades for sites you don't own. Pulse obtained a written exception to these Terms in order to publish grades for the entire federal government.

Also, do you have any interest in turning the guts of this into a `starttls-scan` Python library or CLI tool? That's something @einaros might appreciate, and it would also slim down this scanner into something that just calls out to it.

garrettr:
> Also, do you have any interest in turning the guts of this into a starttls-scan Python library or CLI tool? 

I'm not opposed to that, but don't see an immediate need for that until the existing issues with `starttls.info` are resolved.

garrettr:
@konklone Is this ready to merge, or do I need to follow up on anything first?

scanners/starttls.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def poll_starttls_info_status(domain):
+    """
+    Poll the starttls.info API endpoint until the status of the info for a
+    given `domain` is 'DONE'.
+
+    We poll every 5 seconds.
+
+    # TODO: set a maximum poll time after which we stop trying and report an error
```
konklone: Even if I merge this, this `TODO` about setting up a max poll time is a dealbreaker for me using the scanner -- I get very nervous about `while True` lines.

konklone:
@garrettr Thanks for doing this, and reverse engineering starttls.info. I'm happy to accept, and to see where you take it. Right now, the starttls.info API is a bit too much of a black box (both in source code and documentation) for me to integrate it anywhere publicly, but I'll at least do some private batch scans. I really appreciate it!

PR #39, opened 2015-08-31
Accessibility scan

adelevie:
Close to ready.

Produces a CSV that looks like https://gist.github.com/adelevie/83bb01c024aa05c3b451.

TODOS:
- [ ] Does `PA11Y_PATH` need to be there?
- [ ] Document pa11y dependency requirements. It's all fine in the Dockerfile, but what about the Dockerless?
- [ ] Should we allow the user to alter the scan standard? It currently is hardcoded to WCAG2AA, which is fine for Pulse, but other options are: Section508, WCAG2A, and WCAG2AAA. If so, what's a nice API for that?
- [ ] I'm using some weird thing for getting `stdout` from shelling out to pa11y which skips `util.scan`. I understand this might disrupt Uniform Logging :tm:.

### DIFF
--- a/scanners/a11y.py
+++ b/scanners/a11y.py
@@ -16,7 +16,7 @@ def get_exitcode_stdout_stderr(cmd):
     proc = Popen(args, stdout=PIPE, stderr=PIPE)
     out, err = proc.communicate()
     exitcode = proc.returncode
-    #
+
     return exitcode, out, err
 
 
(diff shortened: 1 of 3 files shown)


test.sh
```diff
@@ -0,0 +1 @@
+PA11Y_PATH=/usr/local/bin/pa11y docker-compose run scan gsa.gov --scan=a11y
```
konklone: Did you mean to include this file in the PR?

adelevie: Kind of sort of ... This _could_ be the beginnings of some simple testing, but I can remove it if that makes things easier.

scanners/a11y.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Execute the external command and get its exitcode, stdout and stderr.
+    """
+    args = shlex.split(cmd)
+
+    proc = Popen(args, stdout=PIPE, stderr=PIPE)
+    out, err = proc.communicate()
+    exitcode = proc.returncode
+
+    return exitcode, out, err
```
konklone: Can we set `stdout` and `stderr` to print out to the usual place?

Do the [arguments in this line](https://github.com/unitedstates/inspectors-general/blob/master/inspectors/utils/utils.py#L299) point in a good direction:

```python
subprocess.Popen(["pdftotext", "-v"], shell=False, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT).communicate()
```

test.sh
konklone: That's fine, you can leave it in.

adelevie:
Rebased.

konklone:
I've pushed up changes in 4f4b556 that fix the stderr issues, add `--level none` to the arguments, implement caching, put the returned JSON in an object instead of an array, and pretty-print the cached results for human readability.

@adelevie Consider what's there reviewed by me.

konklone:
@adelevie What do you think?

konklone:
Closing at @adelevie's suggestion.

PR #41, opened 2015-09-06
Add a local sslyze-based TLS scanner

konklone:
This uses [`sslyze`](https://github.com/nabla-c0d3/sslyze), a solid local TLS analysis tool originally developed by iSEC Labs and now maintained by Alban Diquet, to scan domains' TLS endpoints for key metrics.

### Alternative to SSL Labs

This is meant as an alternative to the already-integrated `tls` scanner, which uses `ssllabs-scan` to hit the SSL Labs API. The SSL Labs API is fantastic -- its methodology is sophisticated, it provides information for a huge range of clients, and its grades are irreplaceable. 

However, SSL Labs' API is also proprietary and (most importantly right now) very slow. Individual SSL Labs scans can take a couple of minutes to perform, and scanning on the order of 300-400 domains takes around 5.5 hours (in part because SSL Labs scans each of the domain's individual endpoints, including IPv6 endpoints). domain-scan's client parallelization can only help so much, as the server vigorously rate-limits clients.

I'm now getting to the point where I want to scan on the order of 10,000 domains or more, which seems like it would take over a week to perform with SSL Labs at current rates. So, at least for the time being, I need a way to more quickly get some key TLS facts about domains.

It's also worth mentioning that SSL Labs can only scan publicly accessible domains. `sslyze` can be used to scan domains only accessible locally.

### Scanner methodology

This scanner requires a path to `sslyze.py`, which should be downloaded from the [sslyze release directory](https://github.com/nabla-c0d3/sslyze/releases) and unzipped somewhere. It can be on the `PATH`, or you can set the `SSLYZE_PATH` env var to point to the location of `sslyze.py`.

Unfortunately, `sslyze` requires Python 2.7, and `domain-scan` requires Python 3, so I'm currently hacking in a solution by specifying a `SSLYZE_PYENV` env var, set to `2.7.9` by default. This causes a `PYENV_VERSION=2.7.9` environment variable to be set during the shell-out to `sslyze.py`. I'd welcome a more generic way to do this.

The effective command this scanner runs is:

```bash
/path/to/sslyze.py --regular --quiet [domain] --xml_out=/path/to/cache/sslyze/[domain].xml
```

The XML output is saved directly from `sslyze`, so the scanner makes sure the directory is present before making this call. Once the command above is run, the XML file is read in, parsed, and a bunch of scalars are plucked out of it. As a convenience, I save a `/path/to/cache/sslyze/[domain].json` file along with the XML, which is the result of parsing the XML.

The base second-level domain is included in the output, along with all the scalars: supported protocol versions, OCSP stapling support, the signature algorithm on the leaf, the key algorithm and length, whether SHA-1 appears in the served chain (discounting roots), whether RC4 appears in any accepted cipher, and the name of the highest served issuer field (an attempt to get the name of the CA, most of the time).

### TODO

- [x] Presence of any accepted forward secret ciphers (ECDHE/DHE)
- [x] Documentation in the README
- [x] Expiration date
- [x] Weakest DH group size
- [x] Fix column mismatch

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,25 +1,35 @@
 ## A domain scanner
 
-Scans domains for data on their:
+Scans domains for data on their HTTPS configuration and assorted other things.
 
-* HTTP, HTTPS, and [HSTS](https://https.cio.gov/hsts/) configuration, using [`site-inspector`](https://github.com/benbalter/site-inspector-ruby).
-* Detailed TLS configuration, using the [SSL Labs API](https://github.com/ssllabs/ssllabs-scan).
-* Whether a domain participates in the [Digital Analytics Program](https://analytics.usa.gov). (This one's U.S. government-specific for now, but can be ignored.)
+**Most of the work is farmed out to other command line tools.** The point of this project is to **coordinate** those tools and produce **consistent data output**.
 
-Most of the work is farmed out to other tools. The point of this scanner is to coordinate those tools and produce consistent data output.
-
-Can be used with any domain, or CSV where domains are the first column, such as the [official .gov domain list](https://github.com/GSA/data/tree/gh-pages/dotgov-domains).
+Can be used with any domain, or CSV where domains are the first column, such as the [official .gov domain list](https://catalog.data.gov/dataset/gov-domains-api-c9856).
 
 ### Requirements
 
-* **Python 3**.
-* `inspect` scanner: **[site-inspector](https://github.com/benbalter/site-inspector)**, version **1.0.2 only**.
-* `tls` scanner: **[ssllabs-scan](https://github.com/ssllabs/ssllabs-scan)**, stable branch.
-* `pageload` scanner: **[phantomas](https://www.npmjs.com/package/phantomas)**, installed through npm.
+The requirements here can be quite diverse, because this tool is just a coordinator for other tools. Communication between tools is handled via CLI and STDOUT.
+
+The overall tool requires **Python 3**. To install dependencies:
+
+```bash
+pip install -r requirements.txt
+```
+
+The individual scanners each require their own dependencies. You only need to have the dependencies installed for the scanners you plan to use.
+
+* `inspect` scanner: **Ruby** and **[site-inspector](https://github.com/benbalter/site-inspector)**, version **1.0.2 only**.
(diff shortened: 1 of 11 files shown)


konklone:
I completed a scan of around 12,000 domains. It took somewhere between 5 and 7 hours (had a false start in there and lost some precision on timing). The only error I got was that apparently the `ocspStapling` field isn't always present, which I fixed in f6e0dac.

Some % of the time is spent on simply XML parsing, querying, and JSON serialization and disk writing, which is done each time. Caching logic controls the actual `sslyze` command and XML disk write, but not the subsequent XML/JSON mechanics. We could potentially save some time by moving the XML and JSON writes into the cached envelope, at the cost of forcing a full network rescan on any change to XML parsing logic. Now that initial development has concluded, that's probably worth it.

konklone:
A quick use of `time` to compare runs of a cached `inspect` scan (sans preload list fetch) with a cached `sslyze` scan suggests that `sslyze` runs 60-70ms slower, suggesting that's the approximate time per-domain to manage the extra XML/JSON/disk mechanics post-cache.

For 12,000 domains, saving 65ms per-domain would only cut 13 minutes off of the 5-7 hour scan, so I don't think it's worth the extra complexity. I'm satisfied with this PR now, and will merge it tomorrow unless other folks watching have issues to raise.

konklone:
I've updated the PR to group `sslyze` information in the dict/JSON more logically, and to allow for errors during certificate parsing. Sometimes `sslyze` throws an exception when certificate parsing, and no `certInfo` is present for analysis. When that happens, the error is logged in a new `Errors` field. I also rearranged the fields to feel more logically grouped.

Finally, I added two more fields: "All Forward Secrecy" and "All RC4" which measure whether **all** accepted ciphersuites for a domain employ `DHE`/`ECDHE` or `RC4`, respectively.

bknowles:
FYI, they recently mentioned the tlsfuzzer project on the oss-security list, which I'm assuming is related to the older sslfuzzer project.

See https://github.com/tomato42/sslfuzzerpython and https://github.com/tomato42/tlsfuzzer

konklone:
Thanks for letting us know! We're measuring server implementations, rather than client implementations, but I wasn't aware of `tlsfuzzer` and I'm definitely glad to.

PR #42, opened 2015-09-08
Always include the base domain

konklone:
This adds a "Base Domain" column to the resulting CSV of every scanner, to the right of the "Domain" column. Now that we're dealing in more subdomains, it's consistently very helpful to be able to sort the results by base domain.


konklone:
Re-opening after merging the PR this was downstream of.

PR #44, opened 2015-09-19
rework dockerfile and add entrypoint functionality

ozzyjohnson:
I wanted to use the docker image to check out collegescorecard and ran into a problem or two. I got it working and added a few things I wanted.

- Pins the version of all package dependencies, very nice to have when doing automated builds.
- Uses RVM instead of Rbenv for the benefit of grabbing a straightforward Ruby binary instead of compiling.
- Adds sslyze 0.11 which seemed to be missing?
- Adds a volume mountpoint at `/data`.
- Adds the `scanner` user for running without root.
- Adds an entrypoint wrapper script with 3 modes.
    - (blank) : Runs everything unchanged (root, output to ./).

         ```
        docker run \
domain-scan \
google.com --scan=inspect,tls,pageload,sslyze
        ```
    - drop : Runs scan as the scanner user.

         ```
        docker run --rm \
domain-scan \
drop \
google.com --scan=inspect,tls,pageload,sslyze
        ```
    - data : Outputs to the `/data` mountpoint, for use with with something like...

        ```
        docker run --rm \
-v /home/ubuntu:/data \
domain-scan \
data \
google.com --scan=inspect,tls,pageload,sslyze
        ```
    ...which will output `cache` and `results` to the host machine `/home/ubuntu` directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,8 +66,6 @@ RUN pip3 install -r requirements.txt
 # Get RVM.
 RUN gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
 RUN curl -sSL https://get.rvm.io | bash -s stable --ruby=2.1.5
-      # && apt-get clean \
-      # && rm -rf /var/lib/apt/lists/*
 RUN /bin/bash -l -c "rvm --default use 2.1.5"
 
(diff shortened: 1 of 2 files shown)


vzvenyach:
Nice work @ozzyjohnson!

konklone:
Yes, super helpful work, thank you!

Issue #49, opened 2015-12-21
Better explanation for https://github.com/18F/domain-scan#setting-tool-paths 


onceuponatimeforever:
The current instructions aren't clear about where/how to set the environmental variables required for localdev.

konklone:
No? I generally feel like environment variables are so standard in Unix environments that I'm comfortable letting people set that in whatever way works for them.

I use [`autoenv`](https://github.com/kennethreitz/autoenv), would adding a mention of that improve the documentation?

onceuponatimeforever:
It's better to be explicit with the installation instructions than assume that new contributors know how to do a thing. 

I'm :+1: to mentioning `autoenv`.

konklone:
@onceuponatimeforever does #50 help?

PR #50, opened 2015-12-23
Add some detail on setting environment variables

konklone:
To address #49 and provide some guidance on setting environment variables.


onceuponatimeforever:
lgtm

PR #63, opened 2016-04-17
Add pyenv to Dockerfile

robbi5:
Since 90665e4ee87f7dfd27b65b4d48278286988e2196 `pyenv` is required for `sslyze` - but the Dockerfile didn't include it yet. This PR hopefully fixes that :)

Includes removal of some pinned versions (commit 4f7e5e91934b24bd137e9125ef36a6bcb0d418b9 of #39)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,12 +19,12 @@ RUN \
         --no-install-suggests \
       build-essential=11.6ubuntu6 \
       curl=7.35.0-1ubuntu2.6 \
-      git=1:1.9.1-1ubuntu0.2 \
+      git \
       libc6-dev=2.19-0ubuntu6.7 \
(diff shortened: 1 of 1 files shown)


konklone:
I'm :+1:, and even though I know there are some reservations about dropping pinning, the maintenance burden makes it seem like the right call. 

@ozzyjohnson @vzvenyach If you have any thoughts, please ring in, otherwise I'll plan to merge this soon.

ozzyjohnson:
I think pulling pinning makes sense. 

Not much point if it's not actively maintained and / or there's no automated build. Also, not essential if building the image isn't in the path of some important workflow.

konklone:
Yeah, it may well be worth revisiting if/when we work this into a production workflow here. I think we will need to before too long.

PR #64, opened 2016-06-02
Docker tweaks

adelevie:
Two changes to get me up and running with Docker:

- Removed a hardcoded version of libc6-dev. Before, `docker-compose build` was breaking at this part. I know it's best practice to have these versions pinned down. But I don't know how to do that. I'll learn it and do it later.

- Use `/home/scanner` as a shared volume with the host. This allows the user to access the results of a scan in the host. Before, you could run scans, but the results were inaccessible, deep inside the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ RUN \
       build-essential=11.6ubuntu6 \
       curl=7.35.0-1ubuntu2.6 \
       git \
-      libc6-dev=2.19-0ubuntu6.7 \
+      libc6-dev \
       libfontconfig1=2.11.0-0ubuntu4.1 \
       libreadline-dev=6.3-4ubuntu2 \
       libssl-dev \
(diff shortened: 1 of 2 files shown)


afeld:
The changes look fine to me 👍

adelevie:
Cool! How would I find the current version of `libc6-dev` so I can re-hardcode it in the Dockerfile?

afeld:
Hmm, that does seem to be the latest version:

http://packages.ubuntu.com/trusty/libc6-dev

> `docker-compose build` was breaking at this part

What was the error?

konklone:
cc @garrettr on this since I know he was having an issue with the second item this PR fixes.

afeld:
Merging this now since it's broken, but feel free to send another PR to lock down the version if you like.

garrettr:
@konklone Sorry, just getting back into working with this now. I can confirm that this PR fixes the issue I was having with retrieving scanner results from the container 👍.

Issue #66, opened 2016-06-02
Suggestion: Split up the Dockerfile


afeld:
Just took a look at the Dockerfile for the first time, and was surprised to see how much is in there. I guess it's because the various tools being used all have different dependencies?

Having multiple languages in a single Dockerfile is an antipattern (IMHO), and I think the setup for each scanner could be a lot simpler if you isolated each tool to its own Dockerfile. These could then be run independently, or via a `domain-scan` Dockerfile that calls out to `docker run <scanner>` and then stitches the results together.

I got this idea from the architecture of [the Code Climate CLI](https://github.com/codeclimate/codeclimate), so you could look there for inspiration if you're interested in pursuing this.

afeld:
Also, reached out to see if the architecture motivation/overview/instructions are written down somewhere:

https://twitter.com/aidanfeldman/status/738376676619505664

konklone:
> Having multiple languages in a single Dockerfile is an antipattern (IMHO), and I think the setup for each scanner could be a lot simpler if you isolated each tool to its own Dockerfile. 

@afeld I agree so very wholeheartedly.

cc @garrettr as well

afeld:
We talked about this suggestion at length [in Slack](https://18f.slack.com/archives/docker/p1464877888000035), but just wanted to give a picture of what the Dockerfiles might look like, to hopefully better demonstrate the idea/benefits:

```dockerfile
# Dockerfile-domain-scan
FROM python:3
CMD glue-code-that-calls-docker-run-on-the-others.py

# Dockerfile-sslyze
FROM python:2
RUN wget ...
CMD sslyze

# Dockerfile-inspect
FROM ruby
RUN bundle install
CMD inspect

...
```

afeld:
Hacked on this a bit tonight: https://github.com/18F/domain-scan/pull/67

konklone:
Closing as inactive.

PR #67, opened 2016-06-04
run (some) scanners in their own Docker containers

afeld:
This pull request switches `ssllabs-scan` and `sslyze` to run in their own Docker containers, rather than in the top-level domain-scan container. Quite happy with how many dependencies this was able to remove from the Dockerfile!

@adelevie Let me know what you think! I'm pretty new to domain-scan, so also please try this locally to make sure it still works as expected.

/cc #66

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,19 +50,6 @@ RUN \
       apparmor \
       docker-engine \
 
-      # Preemptively install these so we don't have to clean up after RVM.
-      autoconf=2.69-6 \
-      automake=1:1.14.1-2ubuntu1 \
-      bison=2:3.0.2.dfsg-2 \
-      gawk=1:4.0.1+dfsg-2.1ubuntu2 \
-      libffi-dev=3.1~rc1+r3.0.13-12ubuntu0.1 \
-      libgdbm-dev=1.8.3-12build1 \
-      libncurses5-dev=5.9+20140118-1ubuntu1 \
-      libsqlite3-dev=3.8.2-1ubuntu2.1 \
-      libtool=2.4.2-1.7ubuntu1 \
-      pkg-config=0.26-1ubuntu4 \
-      sqlite3=3.8.2-1ubuntu2.1 \
-
       # Additional dependencies for python-build
       libbz2-dev \
       llvm \
@@ -71,16 +58,6 @@ RUN \
     # Clean up packages.
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
-
-###
-# Ruby
-
-# Get RVM.
-RUN gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
-RUN curl -sSL https://get.rvm.io | bash -s stable --ruby=2.1.5
-RUN /bin/bash -l -c "rvm --default use 2.1.5"
-
-RUN /bin/bash -l -c "gem install site-inspector -v 1.0.2 --no-ri --no-rdoc"
 
 ###
 # Python dependencies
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -10,13 +10,27 @@ MAINTAINER V. David Zvenyach <vladlen.zvenyach@gsa.gov>
...
+    && echo deb https://apt.dockerproject.org/repo ubuntu-trusty main > /etc/apt/sources.list.d/docker.list \
+    && apt-get update \
+        -qq \
+    && apt-get purge lxc-docker \
+
+    && apt-get install \
+        -qq \
+        --yes \
+        --no-install-recommends \
+        --no-install-suggests \
```
afeld: I had no idea that installing Docker on Ubuntu is such a pain 😒

adelevie:
Well this is pretty interesting!

README.md
```diff
@@ -138,7 +120,6 @@ Some high-priority TODOs here:
 
 * **JSON output**. Refactor scanners to return a dict instead of a row. Have scanners specify both JSON-style field headers *and* CSV-style column headers in a 2-dimensional array. Use this to make it so JSON and CSV can both be serialized with appropriate fields and in the right order. Include JSON results in the `results/` dir.
 * **Handle network loss gracefully.** Right now, the scanner will assume that a domain is "down" if the network is down, and cache that. That makes trusting the results of a batch run iffy. I don't know the best way to distinguish between a domain being unreachable, and the network *making* the domain unreachable.
-* **Upgrade to site-inspector 3.x.** This repo depends on site-inspector 1.0.2, which is behind the times. But, site-inspector 3 needs more testing and work first. site-inspector 2 also is not backwards-compatible, in CLI syntax or in result format.
```
afeld: This isn't quite done yet. I might need some help understanding the intention of what data you want returned.

README.md
```diff
@@ -10,36 +10,18 @@ Can be used with any domain, or CSV where domains are the first column, such as
 
 The requirements here can be quite diverse, because this tool is just a coordinator for other tools. Communication between tools is handled via CLI and STDOUT.
 
-The overall tool requires **Python 3**. To install dependencies:
+The overall tool requires **Python 3** and **Docker**. To install dependencies:
 
 ```bash
 pip install -r requirements.txt
 ```
```
afeld: @adelevie Since the tool now requires Docker for the individual scanners, what do you think about simplifying the instructions to run the `./scan` through Docker exclusively?

adelevie:
Wait, _all_ tools are now in their own separate containers?! Also want to defer to @konklone since he is the one who still is responsible for running all these commands.

README.md
konklone: :+1: from me.

README.md
konklone: It means that this won't work unless site-inspector is locked to 1.0.2.

Where is the Dockerfile versioned that you uploaded to `18fgsa`? It might be nice to version them in this repo.

konklone: Ah, found it: https://github.com/18F/site-inspector-docker I'll fix it there.

afeld: https://github.com/18F/site-inspector-docker/pull/1

afeld:
Would be really awesome if someone wanted to pick up this work 🙏

konklone:
As frustrating as this is for the folks who did the work, I think I'm going to close this. We've since dropped the inspect scanner, reworked the pshtt and sslyze scanners, and simplified the dependency graph.

From an integrity perspective, I'm also uncomfortable relying on multiple 3rd party Docker containers for the dependencies. We could maintain our own Docker container for each one, but I think that would be unwise to take on as a maintenance burden.

I also don't like that this moves `docker run` commands into the scanner code, hardcoding in a Docker dependency to the entire project, even for simple tasks.

So I think I'd prefer to simplify the Dockerfile we have and double down on it. Thanks for working on this, and to be clear, I'm definitely open to hearing alternatives or other arguments for going down this direction.

PR #68, opened 2016-06-17
Unpin libxml2-dev to fix build

garrettr:
`docker-compose build` is broken on master because the pinned version of libxml2-dev is no longer available. Unpinning it allows the build to succeed.

A general note: pinning the apt dependencies seems to cause a lot of maintenance burden. Is it worth it? I generally think that it's the distro's job to keep packages stable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ RUN \
       libreadline-dev=6.3-4ubuntu2 \
       libssl-dev \
       libssl-doc \
-      libxml2-dev=2.9.1+dfsg1-3ubuntu4.7 \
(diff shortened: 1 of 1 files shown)


konklone:
At this point, I'm comfortable removing the pinning. I think a good amount of this is going to change as we move each scanner to its own container, but the pinning update PRs have seemed to be more trouble than the issues we're trying to prevent by pinning so far.

garrettr:
> I think a good amount of this is going to change as we move each scanner to its own container

👍 on this decision. That's being tracked in #67, correct?

konklone:
Yes, it is.

PR #72, opened 2016-07-11
(WIP) Scripts to collect and filter hostnames from various sources

konklone:
This is the start of a small set of scripts (in a separate directory, separate from the `scanners/` system, but still using its `utils.py` methods) oriented around gathering, parsing, and filtering out hostnames that match a given pattern.

I'm using this to harvest `.gov` hostnames from a variety of public sources, such as scans.io, censys.io, CT logs, analytics.usa.gov, and other miscellaneous public sources.

This is a work-in-progress, but opening this here after adding some basic documentation, in case others are interested in this or have suggestions.


konklone:
WIP as it is, the data is being used, and this doesn't conflict with anything else. Merging.

PR #76, opened 2016-08-14
Begins a scanner for the pshtt HTTP inspection tool

konklone:
This creates a scanner that integrates with [`pshtt`](https://github.com/dhs-ncats/pshtt), a Python-based tool. `pshtt` is designed as a single-purpose replacement for the HTTP inspection tool, currently provided via an unmaintained branch of [`site-inspector`](https://github.com/benbalter/site-inspector/tree/erics-mode)) and integrated into domain-scan as the `inspect` scanner.

`pshtt` is a new tool, and is not yet at feature parity with `site-inspector` and the code still needs some more work and refactoring. It outputs a CSV very similar to `inspect.csv`, though the results currently use different logic and it will not reliably give the same results as the `inspect` scanner.

In addition, `pshtt` currently integrates with `sslyze` and requires Python 2, though I would like to see if the use of `sslyze` can be replaced, and if it can be made agnostic to Python 2 or 3. That said, `sslyze` is going to be used somewhere in this tool chain (there's already an `sslyze` scanner in this project), which will require communication between Python 3 and Python 2. It would be useful to figure out what the Right Way is to handle that, because I'm pretty sure that prepending some `pyenv` commands to an unsafe shell execution (as we do now) is not the Right Way.

This is a WIP, but it does work with `pshtt`'s `master` branch at the time I'm filing it, so I think it's merge-ready pretty much any time.

### DIFF
--- a/scanners/ncats.py
+++ b/scanners/ncats.py
@@ -1,0 +1,74 @@
+import logging
+from scanners import utils
+import os
+import json
+
+###
+# == ncats ==
+#
+# Inspect a site's TLS configuration using NCATS' Python site inspector.
+#
+# Currently depends on pyenv to manage calling out to Python2 from Python3.
+###
+
+command = os.environ.get("NCATS_PATH", "site_inspector_cli")
+
+# Kind of a hack for now, other methods of running sslyze with Python 2 welcome
+pyenv_version = os.environ.get("NCATS_PYENV", "2.7.11")
+
+# default to a long timeout
+timeout = 30
+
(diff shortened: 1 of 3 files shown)


konklone:
This PR is ready for merge, once https://github.com/dhs-ncats/pshtt/pull/16 is merged upstream.

konklone:
With https://github.com/dhs-ncats/pshtt/pull/16 patched up and merged, this is good to use in domain-scan.

PR #78, opened 2016-08-22
Fix bug to detect HTTP2 in SSL Labs result

ThibautGery:
I just fixed a small bug 
The  SSL Labs API returns : 
```bash
$ ./ssllabs-scan verif.site
{
    "npnProtocols": "h2 http/1.1"
}
```


konklone:
Thank you!

PR #80, opened 2016-08-23
Pa11y lambda

adelevie:


scripts/a11y/process_a11y.py
```diff
@@ -0,0 +1,144 @@
...
+
+class A11yProcessor(object):
+    ERRORS = {
+        '1_1': 'Alt Tag Errors',
+        '1_3': 'Form Errors',
+        '1_4': 'Color Contrast Errors',
+        '4_1': 'HTML Attribute Errors'
+    }
+
+    BRANCHES = {
```
adelevie: @brendansudol I worry a bit about how this is set up ... I believe we could have many agency names not be listed in this `dict` ... the point is that anything not listed should default to `executive`. cc @konklone

adelevie: Basically, if you can have a default value of `'executive'` if the `dict` lookup is `Null` ... that would be great.

scripts/a11y/process_a11y.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def clean_row(self, row):
+        domain = row[0].lower()
+        agency = self.domain_to_agency.get(domain, 'N/A')
+        code = row[4]
+
+        return {
+            'domain': domain,
+            'agency': agency,
+            'branch': self.agency_to_branch.get(agency, 'Executive'),
```
adelevie: Oh, there it is :)

scripts/a11y/process_a11y.py
```diff
@@ -0,0 +1,144 @@
...
+        return self.ERRORS.get(error_id, 'Other Errors')
+
+    @staticmethod
+    def read_csv(filename):
+        with open(filename, 'r') as f:
+            return [row for row in csv.reader(f)]
+
+
+if __name__ == '__main__':
+    a11y_filename = 'tests/data/a11y.csv'
```
adelevie: I'm guessing these will be overridable?

tests/test_process_a11y.py
```diff
@@ -0,0 +1,47 @@
+import unittest
+
+from scripts.a11y.process_a11y import A11yProcessor
+
+
+class ProcessA11yTestCase(unittest.TestCase):
```
adelevie: wooohooo tests!!!!11

scripts/a11y/process_a11y.py
brendansudol: i believe that's how this works -- this is just a whitelist -- it trys to do the branch lookup (given agency) and if it fails it defaults to executive

scripts/a11y/process_a11y.py
brendansudol: 💥

scripts/a11y/process_a11y.py
brendansudol: yeah, this should be changed to be an overrideable argument (or we can just pick a location and say they have to be there and named properly to run)

brendansudol: the latter is likely sufficient for now -- i'll update paths to use data in a top-level folder called data

scripts/a11y/process_a11y.py
```diff
@@ -134,11 +134,13 @@ def get_error_category(self, code):
...
         with open(filename, 'r') as f:
-            return [row for row in csv.reader(f)]
+            reader = csv.reader(f)
+            next(reader)  # TODO: make header row skip configurable
+            return [row for row in reader]
 
 
 if __name__ == '__main__':
-    a11y_filename = 'results/a11y.csv'
+    a11y_filename = 'data/a11y.csv'
```
adelevie: shouldn't this be `results`? that's where the CSV is saved to once a scan is done.

brendansudol: changed

brendansudol:
@adelevie -- do you feel comfortable addressing these merge conflicts?

adelevie:
>  do you feel comfortable addressing these merge conflicts?
sure

scripts/a11y/process_a11y.py
konklone: Use the `utils.data_dir()` or `utils.cache_dir()` functions to get the right path to `data/` or `cache/` for output or temporary data. These values default to top-level dirs, but will reflect any `--output` override that was passed in to domain-scan.

brendansudol: you can now pass in the datafiles as script args

brendansudol:
i initially merged this into master, but then thought maybe i should maybe to get the A-OK on that from @adelevie or @konklone, so i reverted :)

PR #83, opened 2016-08-28
Like sslyze, have the tls (SSL Labs) scanner use the canonical endpoint

konklone:
At some point, SSL Labs stopped doing its own auto-detection of the canonical endpoint for a domain, and started just scanning whatever endpoint you gave it. (I don't blame them for doing that, given how annoyingly complicated it is.)

This pull request now uses our own canonical endpoint detection to feed the right endpoint to SSL Labs, so that if a site is canonically at `www.domain.gov`, then SSL Labs will scan `https://www.domain.gov` and not `https://domain.gov`.

### DIFF
--- a/scanners/tls.py
+++ b/scanners/tls.py
@@ -25,94 +25,100 @@ def scan(domain, options):
         logging.debug("\tSkipping, HTTPS not supported in inspection.")
         return None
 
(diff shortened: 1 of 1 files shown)


konklone:
Something got screwed up with the merge, I'm trying again.

PR #86, opened 2016-09-14
removing further scans

gbinal:
per https://github.com/18F/pulse/issues/541


scanners/a11y.py
```diff
@@ -63,6 +63,7 @@ def get_errors_from_pa11y_lambda_scan(domain, cache):
                 'WCAG2AA.Principle3.Guideline3_1.3_1_1.H57.3',
                 'WCAG2AA.Principle3.Guideline3_1.3_1_2.H58.1',
                 'WCAG2AA.Principle4.Guideline4_1.4_1_1.F77'
+                'WCAG2AA.Principle4.Guideline4_1.4_1_2.H91'
```
konklone: This will fail, it's missing a comma on the line above. If at all possible, the changes should be tested by someone who :+1:'s it here to say they worked.

PR #94, opened 2016-10-06
Making Lambda optional

adelevie:


gbinal:
Very happy to have this.  This is a blocker for me, so I'm going to merge when the checks are done.  

Thanks.

gbinal:
Sorry, to clarify, I'll hold on merging and just use this branch locally

konklone:
@gbinal Is this still relevant?

gbinal:
Indeed.  Sorry friend.  We'll try to bring it back when we can.  

In the meantime, either @adelevie or @konklone - I know you're both busy but would either of you be okay to give this PR a quick 👀 ?  We think that it's in a place where we can merge it into master but need a 2nd informed (that rules me out) look first.

.gitignore
```diff
@@ -2,6 +2,7 @@
 /data
 *.csv
 *.json
+!pa11y_config.json
```
konklone: Does this really have to live in the project root when running `pa11y`?

pa11y_config.json
```diff
@@ -0,0 +1,61 @@
+{
```
konklone: I'd prefer this file be moved out of the project root, and into a subdir of its own (or some other idea).

scanners/a11y.py
```diff
@@ -6,8 +6,7 @@
 from scanners import utils
 
 
-workers = 25
-PA11Y_STANDARD = 'WCAG2AA'
+workers = 1
```
konklone: Is 1 the desired default here? No parallelization?

scanners/a11y.py
```diff
@@ -48,90 +47,34 @@ def cache_is_not_forced(options):
...
-def get_errors_from_results(results):
-    errors = []
-    for result in results:
-        if result['type'] == 'error':
-            errors.append(result)
-    return errors
+def run_a11y_scan(domain, cache):
+    logging.debug("[%s][a11y]" % domain)
+    pa11y = os.environ.get("PA11Y_PATH", "pa11y")
+    command = [pa11y, domain, "--reporter", "json", "--config", "pa11y_config.json", "--level", "none", "--timeout", "300000"]
```
konklone: This makes me wonder if `pa11y_config.json` should live in the Pulse repo, since that's the project which is setting the configuration and parameters for this scan. You could then pass in a raw.github.com URL to that JSON file to the scan environment, much like various scanners in this project accept either a local or remote URL to content stored elsewhere.

konklone:
@gbinal @micahsaul Sure thing, I just left a review. There are some things I'm hoping we can separate now, to keep domain-scan reasonably (but not unreasonably) separated from Pulse-specific configuration. Thoughts welcome.

scanners/a11y.py
adelevie: Doesn't that unnecessarily complicate things? Keeping `pa11y_config.json` here keeps running this simple for TTS, but I can see where it might complicate things just a little bit for non-TTS users. Maybe keep a `pa11y_config.json.example` checked in here?

konklone: It doesn't seem _so_ complicated -- it'd be pointing to a URL instead of a local path. I guess I'm skeptical that a `pa11y_config.json.example` would be super useful here, since every user of this tool to run `pa11y` scans is going to have their own configuration to manage for some other project, and that config is not likely to look like any particular base template file.

scanners/a11y.py
micahsaul: Hah. Totally inherited that change.

@adelevie, is there a reason not to parallelize this?

adelevie:
Nope. Not sure what the optimal number is for the network-bound processes,
though.

scanners/a11y.py
micahsaul: I'm agnostic here. I'm happy to move the file to its own subdir for now as we decide what we want to do with it long term?

Also, my next project is to move all of the process_a11y.py logic over to Pulse the Pulse repo, so if we decide that's where the config belongs I can include it in that PR.

README.md
```diff
@@ -81,7 +81,7 @@ Parallelization will also cause the resulting domains to be written in an unpred
 * `sslyze` - TLS configuration, using [`sslyze`](https://github.com/nabla-c0d3/sslyze).
 * `analytics` - Participation in an analytics program.
 * `pageload` - Page load and rendering metrics.
-* `a11y` - Accessibility data with the [`pa11y` CLI tool](https://github.com/pa11y/pa11y) via AWS Lambda (requires an AWS account and some additional setup, described further down this document).
+* `a11y` - Accessibility data with the [`pa11y` CLI tool](https://github.com/pa11y/pa11y) (*optionally* via AWS Lambda which, requires an AWS account and some additional setup, described further down this document).
```
micahsaul: Actually, this PR removes the lambda option. @gbinal, game to update the README?

.gitignore
micahsaul: Happy to move it. See my comment on a11y.py

micahsaul:
@konklone Yeah, process_a11y.py and related tests need to be moved into the pulse process.py. I'm doing that as a separate task.

README.md
micahsaul: Done

scanners/a11y.py
micahsaul: Moved it to a config subdirectory. I'm happy to move it to Pulse if that's where we want it, and don't really have an opinion either way.

pa11y_config.json
micahsaul: Done

micahsaul:
I'm going to leave workers=1 for now, but will play around a bit to see if I can find a good number to use there.

micahsaul:
@adelevie, @konklone Game to take another look?

konklone:
Thanks for responding to my requests! Looks good to me.

PR #98, opened 2016-11-20
add pshtt to dockerfile, readme

robbi5:
Currently it's not possible to run the full pulse pipeline with docker, because `pshtt` is missing in the `18fgsa/domain-scan` docker image.

This PR adds support for `pshtt` in the docker images and adds it to the Readme.


konklone:
A+, thank you for doing this!

PR #105, opened 2016-12-27
Bring headers in line with pshtt

h-m-f-t:
This change brings the headers in  line with present-day pshtt.

### DIFF
--- a/scanners/pshtt.py
+++ b/scanners/pshtt.py
@@ -82,7 +82,7 @@ def scan(domain, options):
     yield row
 
 headers = [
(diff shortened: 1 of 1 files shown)


scanners/pshtt.py
```diff
@@ -82,9 +82,10 @@ def scan(domain, options):
     yield row
 
 headers = [
-    "Canonical URL", "Live", "Redirect", "Redirect To",
+    "Domain", "Base Domain", "Canonical URL", "Live", "Redirect", "Redirect To",
```
konklone: These two - `Domain` and `Base Domain` - should not be added to individual scanners. They are added for *all* reports by `domain-scan` automatically.

PR #113, opened 2017-03-10
Fix download URL for sslyze

siccovansas:
The URL where the 0.11 release of sslyze can be downloaded has changed.


konklone:
:+1: Thanks!

PR #114, opened 2017-03-21
Adds a timeout to the pa11y scan

micahsaul:
Some sites are taking >30 minutes, potentially even several hours. This
is untenable, and for now we should just pass over them.


gbinal:
Great, thank you.

PR #116, opened 2017-03-24
Instructions on running an FCPCA-trusting pshtt scan

egyptiankarim:
Added an augmented CA bundle file and instructions to the Dockerfile and README to support running pshtt scans that trust the FCPCA.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ It's possible for scans to save multiple CSV rows per-domain. For example, the `
 
 Example: `results/meta.json`
 
-* **Federal Common Policy CA Trust** to run a Dockerized version of pshtt that uses a FCPCA augmented CA bundle (useful for Federal agencies interested in seeing pshtt status from the perspective of a client footprint that trusts FPKI), uncomment the following two lines in the [Dockerfile](https://github.com/18F/domain-scan/blob/master/Dockerfile):
+* **Federal Common Policy CA Trust**: To run a Dockerized version of pshtt that uses a FCPCA augmented CA bundle (useful for Federal agencies interested in seeing pshtt status from the perspective of a client footprint that trusts the Federal PKI), uncomment the following two lines in the [Dockerfile](Dockerfile):
 
 ```
 #ENV REQUESTS_CA_BUNDLE=$SCANNER_HOME/ssl/requests_plus_fcpca.pem


konklone:
I tweaked the README text and made the link to the Dockerfile relative.

I'd like to talk about the pros/cons of baking a snapshot of the `requests` trust store + FCPCA. It's not the FCPCA part I'm concerned about, and I don't mind optimizing for a dominant use case. It's more that I'm concerned we'll fall behind the `requests` root store, and continue trusting things that have otherwise been (appropriately) yanked from it.

Does `requests` maintain that .pem in source control on GitHub anywhere? Then maybe we could download and cache the root store file as part of the Dockerfile's build process, without baking it in here.

konklone:
Looks like it's here:
https://github.com/kennethreitz/requests/blob/master/requests/cacert.pem

So we could download https://raw.githubusercontent.com/kennethreitz/requests/master/requests/cacert.pem and then also download the latest FCPCA root from its own permalink, and then append the files together before setting them as the common root store.

konklone:
Separately, I also don't love the uncomment strategy, since it could very easily cause you or others to accidentally commit the uncommented changes back to the repo during a subsequent commit (and create hassle for you as you `git stash` or otherwise try to prevent that from happening).

More ideal would be a command line flag given to the build process that opts in to building FCPCA trust into the built container. I'm not super up to speed on how to pass arguments to the Docker build process, but does that sound feasible?

egyptiankarim:
@konklone, you found the link to the Requests repo before I even had a chance to post it!

All your points are very valid, and I was struggling with them myself as I prepared the pull request. 

I'll spend more time looking into how to shove arguments to Docker from the (it must be possible), and look into a smarter way to cat together all the pertinent pems.

konklone:
> I'll spend more time looking into how to shove arguments to Docker (it must be possible), and look into a smarter way to cat together all the pertinent pems.

@egyptiankarim Just checking in, is this something you ended up doing?

konklone:
@egyptiankarim I think once merged, this is going to be the way to use one's own CA bundle in `pshtt` (and `domain-scan`):

https://github.com/dhs-ncats/pshtt/pull/83

Want to take a look, and either close this PR or adapt it to that approach? (Assume it'll be merged.)

egyptiankarim:
@konklone, I took a look at @amirian28's work on dhs-ncats/pshtt#83, and I think we can safely close this PR out. Thanks for the heads up!

konklone:
:+1: Awesome!

konklone:
I'll note that I think it will still take a `domain-scan` modification to pass along the command line flag to the underlying `pshtt` call -- could just be an env var which the `pshtt` scanner knows to look for and then serve up as the appropriate command line flag.

PR #120, opened 2017-04-10
Workaround for cached domains with no errors

micahsaul:
This is indicative of the larger need to refactor and clean up the a11y
scanner. Filing a separate issue for this.


gbinal:
Merging this.

PR #122, opened 2017-05-12
Python settings were breaking the dockerfile

micahsaul:
Dockerfile works again!


konklone:
Are you using Docker locally to run scans?

micahsaul:
@konklone Yup.

PR #123, opened 2017-05-15
Update to sslyze 1.X

konklone:
This updates to sslyze 1.X, which has a new stable Python API, **works in Python 3**, and a revamped approach to certificate parsing.

Unfortunately, I'm not easily able to take full advantage of the Python API while retaining the ability to cache all fetched network data to disk to replay scans, because the Python API doesn't have an easy way to serialize (to JSON or XML) the full results of the scan from within that interface. 

So, I continue to shell out to calling `sslyze`, but this is improved in two ways:

* It drops the use of pyenv, since we're now not depending on Python 2.x. So this also now uses the `utils.scan()` command instead of `utils.unsafe_execute()`.
* It uses JSON as a serialization format instead of XML, from sslyze's `--json_out` flag.

sslyze now doesn't parse certificate info for you in the JSON or XML output, instead asking relying parties to parse the x.509 certificates themselves. So, this makes use of the `cryptography` module to parse PEMs directly.

Some other notes:

* I dropped the OCSP stapling field, since it's harder to get and not as important.
* I added a `Highest Constructed Issuer` field, since sslyze now gives you the full constructed chain -- **if** it was able to verify the chain to one of its included trust stores.
* There's more I could do with time -- the data is richer and better than ever before. This is mostly just trying to be a drop-in replacement for now.

### DIFF
--- a/scanners/sslyze.py
+++ b/scanners/sslyze.py
@@ -222,9 +222,9 @@ def parse_sslyze(raw_json):
         leaf = parse_cert(served_chain[0])
         leaf_key = leaf.public_key()
 
-        if has_attr(leaf_key, "key_size"):
+        if hasattr(leaf_key, "key_size"):
             data['certs']['key_length'] = leaf_key.key_size
-        elif has_attr(leaf_key, "curve"):
+        elif hasattr(leaf_key, "curve"):
             data['certs']['key_length'] = leaf_key.curve.key_size
         else:
             data['certs']['key_length'] = None
--- a/scanners/sslyze.py
+++ b/scanners/sslyze.py
@@ -222,7 +222,12 @@ def parse_sslyze(raw_json):
         leaf = parse_cert(served_chain[0])
         leaf_key = leaf.public_key()
 
-        data['certs']['key_length'] = leaf_key.key_size
+        if has_attr(leaf_key, "key_size"):
+            data['certs']['key_length'] = leaf_key.key_size
+        elif has_attr(leaf_key, "curve"):
+            data['certs']['key_length'] = leaf_key.curve.key_size
+        else:
+            data['certs']['key_length'] = None
 
         if leaf_key.__class__ == cryptography.hazmat.backends.openssl.rsa._RSAPublicKey:
             leaf_key_type = "RSA"
--- a/scanners/sslyze.py
+++ b/scanners/sslyze.py
@@ -97,7 +97,9 @@ def scan(domain, options):
         data['config'].get('any_rc4'), data['config'].get('all_rc4'),
 
         data['certs'].get('key_type'), data['certs'].get('key_length'),
-        data['certs'].get('leaf_signature'), data['certs'].get('any_sha1'),
+        data['certs'].get('leaf_signature'),
+        data['certs'].get('any_sha1_served'),
+        data['certs'].get('any_sha1_constructed'),
         data['certs'].get('not_before'), data['certs'].get('not_after'),
         data['certs'].get('served_issuer'), data['certs'].get('constructed_issuer'),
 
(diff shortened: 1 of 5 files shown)


scanners/sslyze.py
```diff
@@ -180,62 +180,91 @@ def parse_sslyze(xml):
...
         if weakest_dh == 1234567890:
             weakest_dh = None
 
         data['config']['weakest_dh'] = weakest_dh
 
     # If there was an exception parsing the certificate, catch it before fetching cert info.
-    if target.select_one("certinfo[exception]") is not None:
-        data['errors'] = target.select_one("certinfo[exception]")["exception"]
+    if False:
+        data['errors'] = "TODO"
```
konklone: I'm not yet sure whether this will ever happen in the new sslyze code path. It'll become obvious in the output if it does!

scanners/sslyze.py
```diff
@@ -180,62 +180,91 @@ def parse_sslyze(xml):
...
+        leaf = parse_cert(served_chain[0])
+        leaf_key = leaf.public_key()
 
-        # Key algorithm and length for leaf certificate only.
-        data['certs']['key_type'] = leaf.select_one("subjectPublicKeyInfo publicKeyAlgorithm").text
-        data['certs']['key_length'] = int(leaf.select_one("subjectPublicKeyInfo publicKeySize").text)
+        if hasattr(leaf_key, "key_size"):
+            data['certs']['key_length'] = leaf_key.key_size
+        elif hasattr(leaf_key, "curve"):
+            data['certs']['key_length'] = leaf_key.curve.key_size
```
konklone: Filed https://github.com/pyca/cryptography/issues/3540 to track the issue that, if fixed, would remove the need for this ugly workaround.

scanners/sslyze.py
```diff
@@ -180,62 +180,91 @@ def parse_sslyze(xml):
...
+        data['certs']['not_after'] = leaf.not_valid_after
 
         # Look at only served leaf and intermediate certificates
-        signatures = target.select("certificateChain certificate[position=leaf],certificate[position=intermediate] signatureAlgorithm")
-        any_sha1 = False
-        for signature in signatures:
-            if "sha1With" in signature.text:
-                any_sha1 = True
-        data['certs']['any_sha1'] = any_sha1
+        data['certs']['any_sha1'] = target['certinfo']['has_sha1_in_certificate_chain']
```
konklone: Note that I'm now relying on sslyze's internal check to see if SHA-1 is anywhere in the chain. This may change it to the constructed chain, where before it was the served chain -- I am not sure, and don't care enough to figure it out.

scanners/sslyze.py
```diff
@@ -180,62 +180,91 @@ def parse_sslyze(xml):
...
 
     return data
+
+# Given the cert sub-obj from the sslyze JSON, use
+# the cryptography module to parse its PEM contents.
+def parse_cert(cert):
+    backend = cryptography.hazmat.backends.openssl.backend
```
konklone: I'm locking it to the openssl backend, since right now I have to introspect on exact class name to distinguish between RSA, DSA, and ECDSA keys. I'm going to file an issue upstream to ask about a way out of that kludge.

konklone: Filed https://github.com/pyca/cryptography/issues/3541

scanners/sslyze.py
konklone: It's from the constructed chain. When there's no constructed chain, this value is blank, and so no conclusion about SHA-1 in the served chain is drawn. I think we probably want two columns - one for whether SHA-1 was in the served chain, and one for whether it was in the constructed chain.

PR #124, opened 2017-05-21
Gathering from multiple sources at once

konklone:
This pull request adds a new way of gathering hostnames from multiple sources at once. Results are collectively deduped, and the resulting CSV will show whether each hostname was found in each checked source.

An example (real-world) use of this facility, to gather publicly discoverable federal .gov hostnames from Censys.io, the Digital Analytics Program, the 2016 End of Term Archive, and the official .gov domain list:

```bash
./gather censys,dap,eot2016,parents --eot2016=https://github.com/GSA/data/raw/gh-pages/end-of-term-archive-csv/eot-2016-seeds.csv --dap=https://analytics.usa.gov/data/live/sites-extended.csv --suffix=.gov --parents=https://raw.githubusercontent.com/GSA/data/gh-pages/dotgov-domains/current-federal.csv --include-parents --debug --export
```
As shown in this example, simple URL/file-based gatherers are dynamically created by giving them a name that isn't already defined as a gatherer (right now, anything other than `censys`), and then specified by using a command line flag with the same name. This dynamic approach is used to keep syntax simple and avoid having to use more complicated syntax (such as numerically sequenced names), while allowing multiple URL/file-based gatherers to run separately without their namespaces colliding.

### DIFF
--- a/gather
+++ b/gather
@@ -53,16 +53,24 @@ def run(options=None):
     hostnames_cache = {}
 
     for source in sources:
+        extra = {}
 
         try:
             gatherer = importlib.import_module("gatherers.%s" % source)
         except ImportError:
-            exc_type, exc_value, exc_traceback = sys.exc_info()
-            logging.error("[%s] Gatherer not found, or had an error during loading.\n\tERROR: %s\n\t%s" % (source, exc_type, exc_value))
-            exit(1)
+            # If it's not a registered module, allow it to be "hot registered"
+            # as long as the user gave us a flag with that name that can be
+            # used as the --url option to the URL module.
+            if options.get(source):
+                gatherer = importlib.import_module("gatherers.url")
+                extra['name'] = source
+            else:
+                exc_type, exc_value, exc_traceback = sys.exc_info()
+                logging.error("[%s] Gatherer not found, or had an error during loading.\n\tERROR: %s\n\t%s" % (source, exc_type, exc_value))
+                exit(1)
 
         # Iterate over each hostname.
-        for domain in gatherer.gather(suffix, options):
+        for domain in gatherer.gather(suffix, options, extra):
 
             # Always apply the suffix to returned names.
             if not suffix_pattern.search(domain):
@@ -106,6 +114,7 @@ def run(options=None):
     hostnames.sort()
 
     for hostname in hostnames:
+        base = utils.base_domain_for(hostname)
         row = [hostname, base]
         for source in sources:
             row += [source in hostnames_cache[hostname]]
--- a/gather
+++ b/gather
@@ -48,11 +48,6 @@ def run(options=None):
     else:
         filename = sources[0] # backwards compatibility, mostly
 
-    gathered_filename = "%s/%s.csv" % (utils.results_dir(), filename)
-    gathered_file = open(gathered_filename, 'w', newline='')
-    gathered_writer = csv.writer(gathered_file)
-    gathered_writer.writerow(["Domain", "Base Domain", "Source"])
-
     # De-duping hostnames. This will cause the system to hold all
     # hostnames in memory at once, but oh well.
     hostnames_cache = {}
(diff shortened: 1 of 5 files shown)


konklone:
cc @h-m-f-t if you have any thoughts

gather
```diff
@@ -49,41 +39,86 @@ def run(options=None):
...
+
     # Open CSV writer.
-    gathered_filename = "%s/%s.csv" % (utils.results_dir(), source)
-    gathered_file = open(gathered_filename, 'w', newline='')
-    gathered_writer = csv.writer(gathered_file)
-    gathered_writer.writerow(["Domain", "Base Domain"])
+    if len(sources) > 1:
+        filename = "gathered"
+    else:
+        filename = sources[0] # backwards compatibility, mostly
```
konklone: For backwards compatibility, if only one source is used then the filename will be named after that source (e.g. `censys.csv`), and if multiple are used then the filename will be `gathered.csv`.

gather
```diff
@@ -49,41 +39,86 @@ def run(options=None):
...
+                if base not in parents:
+                    continue
 
-        # Use hostname cache to de-dupe, if seen before.
-        if domain not in hostnames_cache:
+            # Use hostname cache to de-dupe, if seen before.
+            if domain not in hostnames_cache:
+                hostnames_cache[domain] = [source]
+            elif source not in hostnames_cache[domain]:
+                hostnames_cache[domain] += [source]
```
konklone: The `hostnames_cache` dict is now not only used to dedupe which hostnames appear, but also to track which sources each hostname does and doesn't appear in.

gather
```diff
@@ -49,41 +39,86 @@ def run(options=None):
...
+
+    # Open CSV file.
+    gathered_filename = "%s/%s.csv" % (utils.results_dir(), filename)
+    gathered_file = open(gathered_filename, 'w', newline='')
+    gathered_writer = csv.writer(gathered_file)
+    gathered_writer.writerow(headers)
+
+    # Write each hostname to disk, with all discovered sources.
+    hostnames = list(hostnames_cache.keys())
+    hostnames.sort()
```
konklone: Hostnames are always all sorted in-memory before being output -- otherwise the resulting output would be too unpredictable. But this does mean that super-large gathering jobs might become unwieldy.

gatherers/censys.py
```diff
@@ -49,7 +49,7 @@
 wildcard_pattern = re.compile("^\*\.")
 redacted_pattern = re.compile("^(\?\.)+")
 
-def gather(suffix, options):
+def gather(suffix, options, extra={}):
```
konklone: The ability to use the `url` gatherer under different names meant changing the method signature for each gatherer's `gather` function to include an optional `extra` dict, which might have information that doesn't appear directly in the command line flags. 

In this case, the `url` gatherer's `gather` method has to know, at call-time, which of the names it's operating under, so that it can pull the right CSV URL/path from the right command line flag.

scanners/sslyze.py
```diff
@@ -250,11 +251,11 @@ def parse_sslyze(raw_json):
...
             data['certs']['key_length'] = None
 
-        if leaf_key.__class__ == cryptography.hazmat.backends.openssl.rsa._RSAPublicKey:
+        if isinstance(leaf_key, rsa.RSAPublicKey):
             leaf_key_type = "RSA"
-        elif leaf_key.__class__ == cryptography.hazmat.backends.openssl.dsa._DSAPublicKey:
+        elif isinstance(leaf_key, dsa.DSAPublicKey):
             leaf_key_type = "DSA"
-        elif leaf_key.__class__ == cryptography.hazmat.backends.openssl.ec._EllipticCurvePublicKey:
+        elif isinstance(leaf_key, ec.EllipticCurvePublicKey):
```
konklone: This change wasn't related to multiple gatherers and probably should be in its own PR, but it also doesn't change behavior -- it's addressing the advice in https://github.com/pyca/cryptography/issues/3541#issuecomment-301367551 as to how to properly introspect the key type of a certificate object returned by the Python `cryptography` module.

PR #125, opened 2017-05-22
Allow specifying a custom Censys.io query

moschlar:


konklone:
This seems like a :+1: to me, though could you share what query you're running? Depending on what it is, it might make sense to support it directly within the tool.

moschlar:
I used the query `parsed.names: uni-mainz.de` because with the search turned up no certificate at all when using the default query with `--suffix=uni-mainz.de`.

konklone:
Is that because you're trying to bring up email certificates, or is this for hostnames of internet/web services?

moschlar:
It's really for finding all the hostnames below our university top-level domain.
You can easily try it yourself - the search query based on just the suffix yields no results...

gbinal:
I'm cool with it.  @micahsaul, @konklone - anything you want before we merge?

micahsaul:
LGTM, but defer to @konklone

gatherers/censys.py
```diff
@@ -79,7 +79,10 @@ def paginated_mode(suffix, options, uid, api_key):
 
     certificate_api = certificates.CensysCertificates(uid, api_key)
 
-    query = "parsed.subject.common_name:\"%s\" or parsed.extensions.subject_alt_name.dns_names:\"%s\"" % (suffix, suffix)
+    if 'query' in options and options['query']:
```
konklone: A small nit - this can be `if options.get('query'):` (it's a safe function call, and will return `None` if `query` isn't a key in the dict). But I can make that change myself, no reason to hold up the query.

PR #131, opened 2017-06-17
Cache preload list across individual runs of pshtt, not domain-scan

konklone:
`domain-scan` sets the `--preload-cache` option for `pshtt` only to keep the preload list JSON cached across each individual run of `pshtt`, to avoid constantly re-downloading it.

However, we weren't flushing this cache in any way between runs of `domain-scan`, so any repeated runs of `domain-scan` in the same directory were reusing the same `preload-list.json` file ad infinitum.

This PR automatically deletes any cached `preload-list.json` file before the `pshtt` scanner is initialized, which ensures that a fresh one will be used for any batch of `pshtt` calls.

Fixes #707.

### DIFF
--- a/scanners/pshtt.py
+++ b/scanners/pshtt.py
@@ -27,6 +27,16 @@ user_agent = os.environ.get("PSHTT_USER_AGENT", "github.com/18f/domain-scan, psh
 preload_cache = utils.cache_single("preload-list.json")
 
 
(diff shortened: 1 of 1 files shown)


konklone:
cc @jamcat22 this should fix the issue you reported, once I merge and deploy to production.

PR #139, opened 2017-07-28
Prepare a11y for automation

micahsaul:
Fixed the non-live and redirect logic.
Blacklist some sites, and manually config redirects for certain sites that use META and JS redirects.


config/a11y-redirects.yml
```diff
@@ -0,0 +1,63 @@
...
+  blacklist: false
+stopfakes.gov:
+  redirect: https://www.stopfakes.gov/welcome
+  blacklist: false
+sciencebase.gov:
+  redirect: https://www.sciencebase.gov/catalog/
+  blacklist: false
+realestatesales.gov:
+  redirect: https://realestatesales.gov/gsaauctions/gsaauctions/
+  blacklist: false
```
konklone: @micahsaul This new .gov-specific YAML file strikes me as something better placed in the Pulse repo, and then pulled in (or downloaded as necessary) by the domain-scan scanner. Other scanners do similar things with data files stored in external GitHub repos or other URLs.

konklone: Is this all to support following client-side redirects? I would honestly rather build in support for following client-side redirects (in either `pshtt` or `domain-scan`) than start hard-coding a giant list of client-side redirects. These are bound to grow stale, and following client-side redirects is valuable for other non-a11y things.

gbinal: > I would honestly rather build in support for following client-side redirects...

@konklone - Totally appreciate that but I'm not sure if we have the bandwidth to do that at this stage and it's clear if/when we will.  I definitely don't want to do anything that imposes on the other scans on the pulse, but we're necessarily not in a position to optimize the wider functionality right now.

gbinal: @micahsaul - any sense on the LOE for moving the YAML file into the Pulse repo?

micahsaul: I agree it'd be nice to have support for following client-side redirects, and this is definitely a (pretty ugly) hack to get around that. That said, that functionality seems like it should be in pshtt and I'm not sure I have the bandwidth to both learn how pshtt is structured and add in meta and js redirect following.

Is the desire to have this live in Pulse instead of here based on wanting .gov/pulse specific config in pulse so domain-scan is corpus agnostic? If so, that makes sense to me and I am happy to move both the a11y config files to Pulse.

konklone: > I agree it'd be nice to have support for following client-side redirects, and this is definitely a (pretty ugly) hack to get around that. That said, that functionality seems like it should be in pshtt and I'm not sure I have the bandwidth to both learn how pshtt is structured and add in meta and js redirect following.

Yep, I understand, it's non-trivial (and that's why it hasn't been implemented yet).

> Is the desire to have this live in Pulse instead of here based on wanting .gov/pulse specific config in pulse so domain-scan is corpus agnostic? 

Yep, that's exactly right.

micahsaul: Moving config to pulse here: https://github.com/18F/pulse/pull/717
I'll switch this over to looking for it there.

konklone: Merged 18f/pulse#717.

micahsaul:
Thanks @konklone! I've added optional command line flags for the a11y config files. I think with that we should be good to go on this.

Dockerfile
```diff
@@ -136,7 +136,7 @@ RUN npm install \
 ###
 # pshtt
 
-RUN pip3 install pshtt
+RUN pip3 install pshtt==0.2.1
```
konklone: A `pip install pshtt` failed?

cc @h-m-f-t

config/a11y-redirects.yml
konklone: @micahsaul I merged, but just saw the config file is still in this repo - can it be deleted?

micahsaul:
Thanks @konklone!

PR #140, opened 2017-08-03
analytics reporter link added to README.md

EricSchles:
Just being explicit about what the `analytics` command is.  So I linked out to analytics-reporter.


konklone:
domain-scan doesn't use the `analytics` command, and that part of the documentation is referring to participation in any analytics program, not necessarily one that has anything to do with `analytics-reporter`. So I think the linkage here is not accurate.

(Though honestly, it might be best just to remove that from the public documentation anyhow, since it's so USG-specific.)

PR #145, opened 2017-10-04
Add new fields from pshtt

h-m-f-t:
"HTTPS Self Signed Cert”, and “Unknown Error” were added to pshtt in the last few months ([here](https://github.com/dhs-ncats/pshtt/commit/2c33255f856b64fc4cc367c82b809ef0befc5a52) and [here](https://github.com/dhs-ncats/pshtt/commit/129f112ec3390da17ed61fd1d05477897708572c), respectively). This change allows these fields to be represented in domain-scan scans.


konklone:
:+1: Thanks!

PR #146, opened 2017-10-04
Added HSTS to sslyze and tls

grgras71:


konklone:
Hi @grgras71, our preferred way to get HSTS information is via `pshtt`, which has a scanner in this project. 

The `pshtt` scanner is smarter than the `sslyze` or `tls` scanners can be, because the `pshtt` scanner factors in the behavior of different endpoints for a domain. This is especially relevant for preloading and the inclusion of subdomains.

For example, if you scan `www.google.com` with SSLyze and you see an HSTS header with a `preload` directive, that doesn't mean `google.com` is preloaded. It has to be there at `https://google.com`.

Using `pshtt` can also improve the results of using `sslyze` or `tls`, because `pshtt` will attempt to determine the "canonical" endpoint for a particular domain, and the `sslyze` and `tls` scanners will use this information if present. So for example, you can scan a list of domain names, and have `sslyze` ping the most correct endpoint for each domain (`https://example.com` or `https://www.example.com`) as appropriate for each one.

So, I'm inclined to close this in favor of relying on the `pshtt` scanner. Will that work for your use case?

PR #147, opened 2017-10-25
Add a column for detecting any use of 3DES

konklone:
This examines OpenSSL ciphersuites for mention of `3DES` specifically, or `DES-CBC3`, which appears to reliably indicate 3DES in CBC mode.

cc @h-m-f-t

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 
 echo "running linters and tests..."
-flake8 . && python -m unittest discover tests
+flake8 --ignore=E722,E501 . && python -m unittest discover tests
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 
 echo "running linters and tests..."
-flake8 . && python -m unittest discover tests
+flake8 --ignore=E722,E501 . && python -m unittest discover tests
```
IanLee1521: Might be preferable to move these into a top level `.flake8` file like we have in pshtt: https://github.com/dhs-ncats/pshtt/blob/master/.flake8

That way if `flake8` is called manually or via some other method, you get the same ignores / results.

konklone: Actually, we already had one! I just forgot about it entirely. Fixed in https://github.com/18F/domain-scan/commit/994cb3892f42be4f91b3555152d8cf2b41974cc2, thanks for catching that.

PR #148, opened 2017-10-29
Added caching for public suffix list for pshtt

jsf9k:
The [public suffix list](https://publicsuffix.org/), which is only updated weekly, was being downloaded for each instance of `pshtt` that `domain-scan` was kicking off.  This change caches the public suffix list data, very similarly to the way the preload data is cached.


konklone:
Thank you! It's been a TODO for a while to rationalize pshtt's caching to have all third party services it touches get cached in a single location, which then pshtt (and other tools) could manage in a simpler way. This is definitely a useful fix, and also useful to understand that this was a concrete source of pshtt hanging. It's possible something similar happens with hstspreload.org.

PR #149, opened 2017-10-30
Correct load_domains behavior for skipping rows

ericlaw1979:
Ignore empty lines and only ignore a "header" if it's the first entry in the file.


ericlaw1979:
It occurs to me that this pull is probably supposed to be over at https://github.com/dhs-ncats/pshtt/pull/129

konklone:
@ericlaw1979 It helps in both places! Thanks for the submission.

Issue #138, opened 2017-07-01
sslyze appearing to stall out on some domains


konklone:
I don't have enough information to report a bug yet. When I checked on our server, the sslyze scans had stalled out with 9 in-flight, with a bunch of `defunct` sslyze processes. These were the domains:

```
nces.ed.gov
autodiscover.ors.od.nih.gov
stg-reg2.hcia.cms.gov
vpn1.cjis.gov
portcullis.nlrb.gov
safesupportivelearning.ed.gov
my.uscis.gov
www.educationusa.state.gov
pittsburgh.feb.gov
```

But when I ran a scan using `sslyze` on all 9 of those in a row, using `--serial`, none of them stalled out. So I'm not totally sure how to reproduce this.

konklone:
cc @h-m-f-t if you've seen anything like this in your own work

h-m-f-t:
```
portcullis.nlrb.gov
my.uscis.gov
```
These two look familiar; I'm pretty certain they threw occasional errors in the early days of pshtt (which probably would have relied on now-defunct capabilities of sslyze). I don't think anything was done client-side to remediate, other than a later re-scan.

jsf9k:
@konklone and @h-m-f-t, I submitted #148 because I was seeing something like this with `pshtt` scans.  When running lots of `pshtt` scans via `domain-scan` I would every once in a while see one or more scans hang indefinitely.  I was able to attach to the hung processes using `gdb` and the debug version of `python`.  The hanging _seems_ to be related to the [public suffix data](https://publicsuffix.org/) being being pulled for every `pshtt` scan being run.  I guess if you pull too often it starts hanging.

I don't know if something similar could be happening inside of `sslyze` runs.  The problem I was seeing would happen with different domains each time, so it was definitely not particular to any given domain.

konklone:
I think I have a decent fix for this in #138. While it doesn't totally identify the root cause, it refactors the tool to use the sslyze library directly, in Python, which cuts down our level of process spawning and indirection. Early signs show that it appears to be making a significant impact on reliability and mitigating zombie processes.

jsf9k:
@konklone, I learned recently that `gdb` can walk python stack traces.  I don't know what OS you're using, but on most linuxes you install the debug version of python, and then run `gdb /path/to/debug/python PID_of_process_to_be_debugged`.  At that point you should be able to run the `py-*` functions like `py-bt` and `py-list` at the `gdb` prompt.

[...]

PR #150, opened 2017-10-31
Local timeout that kills sslyze scans -- but doesn't clean up processes

konklone:
This is a partial fix for #138, that adds a local timeout library that successfully cuts off sslyze processes that get zombified for unknown reasons. 

The full solution is to migrate our use of sslyze to the Python API, now that sslyze works on Python 3 and has such an API. This would also help us track down the source of the hangup in a more granular way, as opposed to the process being a black box. 

We'd likely still also benefit from this work on a local timeout, and it could be useful for other scanners, so this seems worth doing.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ pyyaml
 # to support sslyze scanner
 sslyze
 cryptography
+timeout-decorator
 
 # to support censys gatherer
 censys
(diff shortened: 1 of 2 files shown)


konklone:
To make this more explicit: if sslyze processes get zombied, I _believe_ they may end up needing a manual `kill` sweep to clean them up, but the scan can continue. That's why this is only partial.

konklone:
Closed in favor of #151.

PR #151, opened 2017-11-01
Using SSLyze as a Python library instead of shelling out

konklone:
This is a better approach than #150, which moves us to using sslyze as a Python library rather than shelling out.

It still uses the concurrent scanner, which itself does spin out multiple processes, so this may yet get us into trouble, but it is now easy for us to use the synchronous scanner if we want to, and in general have more control over the parsing experience. 

The timeout library I introduced in #150 can't function with this, as the sslyze native Python calls to do its own multiprocess work interfere with the timeout's multiprocess work when signals are disabled. (When signals are enabled, the timeout library throws an error about sslyze's use of multiple processes. There's no winning.)

If this fails to fix the hanging issues, I'll switch it to use the SynchronousScanner, which will eliminate all use of multiple processes, and we'll get more granular as needed about the root cause.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -153,8 +153,6 @@ RUN pip3 install pshtt
 ENV SCANNER_HOME /home/scanner
 RUN mkdir $SCANNER_HOME
 
-COPY . $SCANNER_HOME
-
 RUN groupadd -r scanner \
   && useradd -r -c "Scanner user" -g scanner scanner \
   && chown -R scanner:scanner ${SCANNER_HOME}
@@ -169,3 +167,5 @@ WORKDIR $SCANNER_HOME
 VOLUME /data
 
 ENTRYPOINT ["./scan_wrap.sh"]
+
+COPY . $SCANNER_HOME
--- a/Dockerfile
+++ b/Dockerfile
@@ -125,16 +125,6 @@ ENV PATH /go/bin:$PATH
 # Node
 RUN ln -s /usr/bin/nodejs /usr/bin/node
 
-###
-# ssllabs-scan
-
-RUN mkdir -p /go/src /go/bin \
-      && chmod -R 777 /go
-RUN go get github.com/ssllabs/ssllabs-scan
-RUN cd /go/src/github.com/ssllabs/ssllabs-scan/ \
-      && git checkout stable \
-      && go install
-ENV SSLLABS_PATH /go/bin/ssllabs-scan
 
 ###
 # phantomas
@@ -149,7 +139,12 @@ RUN npm install \
 ###
 # pshtt
 
-RUN pip3 install pshtt==0.2.1
+RUN apt-get install -qq --yes locales
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
+RUN pip3 install pshtt
 
 
 ###
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ pyyaml
 # to support sslyze scanner
 sslyze
 cryptography
+timeout-decorator
 
 # to support censys gatherer
 censys
--- a/scanners/pshtt.py
+++ b/scanners/pshtt.py
@@ -50,7 +50,7 @@ def scan(domain, options):
 
     if (force is False) and (os.path.exists(cache_pshtt)):
         logging.debug("\tCached.")
-        raw = open(cache_pshtt).read()
+        raw = utils.read(cache_pshtt)
         data = json.loads(raw)
         if (data.__class__ is dict) and data.get('invalid'):
             return None
--- a/scanners/sslyze.py
+++ b/scanners/sslyze.py
@@ -4,9 +4,15 @@ import os
 
 import timeout_decorator
 
+import sslyze
+from sslyze.concurrent_scanner import ConcurrentScanner, PluginRaisedExceptionScanResult
+from sslyze.plugins.openssl_cipher_suites_plugin import Tlsv10ScanCommand, Tlsv11ScanCommand, Tlsv12ScanCommand, Sslv20ScanCommand, Sslv30ScanCommand
+from sslyze.plugins.certificate_info_plugin import CertificateInfoScanCommand
+
 import json
 import cryptography
 import cryptography.hazmat.backends.openssl
+from cryptography.hazmat.primitives.serialization import Encoding
 from cryptography.hazmat.primitives.asymmetric import ec, dsa, rsa
 
 ###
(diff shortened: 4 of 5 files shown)


konklone:
Not for sure yet, but I've been watching somewhere between 30-60 minutes of scans roll by, and no defunct sslyze processes yet! They were common and quick to develop when using the existing method of shelling out. Very promising.

konklone:
I added an update in https://github.com/18F/domain-scan/issues/138#issuecomment-341942822 - this PR now has both synchronous and asynchronous modes (synchronous can be turned on with `--sslyze-serial`), but I'm constantly running out of memory. I believe this was happening before as well with sslyze, which was causing some processes to get killed and become zombies. In synchronous mode here, no processes get spawned, and so the kernel kills the main process entirely.

Watching the memory increase over the lifetime of a scan, it looks like a memory leak is happening, and probably was happening previously. I'm going to try to track it down before merge, even though it was likely happening before this PR as well.

konklone:
Well, after trying a bunch of things, and the most effective thing I found was to 1) use the **concurrent** scanner, and 2) add a network_timeout to the server connectivity call. 

I've added options to try various combinations of serializing sslyze's scans, and disabling certificate info scans (since they are currently informational and not relevant to M-15-13 or BOD 18-01). A memory leak is consistently reproducible with the synchronous scanner, but not with the concurrent scanner.

It's been a wild ride, but I'm becoming comfortable with this PR, and plan to fix the build and merge it in shortly.

requirements.txt
```diff
@@ -6,6 +6,7 @@ pyyaml
 # to support sslyze scanner
 sslyze
 cryptography
+timeout-decorator
```
konklone: Oops, a leftover from intermediate work -- I removed this in `57b43a1`.

Dockerfile
```diff
@@ -161,3 +167,5 @@ WORKDIR $SCANNER_HOME
 VOLUME /data
 
 ENTRYPOINT ["./scan_wrap.sh"]
+
+COPY . $SCANNER_HOME
```
konklone: Moved the `COPY` command last, since it's non-deterministic, and Docker knows this and will always rebuild from here during `docker build`.

Dockerfile
```diff
@@ -136,7 +139,12 @@ RUN npm install \
 ###
 # pshtt
 
-RUN pip3 install pshtt==0.2.1
+RUN apt-get install -qq --yes locales
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
+RUN pip3 install pshtt
```
konklone: Locked to latest `pshtt`. This also required me to paste in please-just-make-it-work attempts at fixing a Unicode error that was happening because of the PyPi `publicsuffixlist` module's LICENSE file having Unicode characters in the author's name.

Dockerfile
```diff
@@ -112,16 +125,6 @@ ENV PATH /go/bin:$PATH
...
-###
-# ssllabs-scan
-
-RUN mkdir -p /go/src /go/bin \
-      && chmod -R 777 /go
-RUN go get github.com/ssllabs/ssllabs-scan
-RUN cd /go/src/github.com/ssllabs/ssllabs-scan/ \
-      && git checkout stable \
-      && go install
-ENV SSLLABS_PATH /go/bin/ssllabs-scan
```
konklone: The `tls` scanner and the use of `ssllabs-scan` is deprecated.

Dockerfile
```diff
@@ -81,12 +77,29 @@ RUN wget ${PYENV_REPO}/archive/v${PYENV_RELEASE}.zip \
...
-    && pyenv install $PYENV_PYTHON_VERSION \
-    && pyenv local $PYENV_PYTHON_VERSION
+    && pyenv install --debug --keep $PYENV_PYTHON_VERSION \
+    && pyenv local ${PYENV_PYTHON_VERSION}-debug
+RUN ln -s /opt/pyenv/sources/${PYENV_PYTHON_VERSION}-debug/Python-${PYENV_PYTHON_VERSION}/python-gdb.py /opt/pyenv/versions/${PYENV_PYTHON_VERSION}-debug/bin/python3.6-gdb.py
+RUN ln -s /opt/pyenv/sources/${PYENV_PYTHON_VERSION}-debug/Python-${PYENV_PYTHON_VERSION}/python-gdb.py /opt/pyenv/versions/${PYENV_PYTHON_VERSION}-debug/bin/python3-gdb.py
+RUN ln -s /opt/pyenv/sources/${PYENV_PYTHON_VERSION}-debug/Python-${PYENV_PYTHON_VERSION}/python-gdb.py /opt/pyenv/versions/${PYENV_PYTHON_VERSION}-debug/bin/python-gdb.py
+RUN apt-get -qq update && \
+    apt-get -qq --yes --no-install-recommends --no-install-suggests install gdb
+RUN echo add-auto-load-safe-path /opt/pyenv/sources/${PYENV_PYTHON_VERSION}-debug/Python-${PYENV_PYTHON_VERSION}/ >> etc/gdb/gdbinit
```
konklone: Thanks to @jsf9k for his work in figuring out how to create a `gdb`-ready version of Python that is installed via pyenv, in the Docker container.

scanners/pshtt.py
```diff
@@ -50,7 +50,7 @@ def scan(domain, options):
 
     if (force is False) and (os.path.exists(cache_pshtt)):
         logging.debug("\tCached.")
-        raw = open(cache_pshtt).read()
+        raw = utils.read(cache_pshtt)
```
konklone: Explicitly closing some not-explicitly-closed file handles, which Docker was yelling about. In a couple places below as well.

scanners/sslyze.py
```diff
@@ -14,8 +21,17 @@
 #
 # If data exists for a domain from `pshtt`, will check results
 # and only process domains with valid HTTPS, or broken chains.
+#
+# Supported options:
+#
+# --sslyze-serial - If set, will use a synchronous (single-threaded
+#   in-process) scanner. Defaults to false.
```
konklone: The `--sslyze-serial` option isn't documented in the README, it was just used for debugging. In the future, I may remove the SSLyze serial code path altogether, as it generates a memory leak under normal use.

scanners/sslyze.py
```diff
@@ -14,8 +21,17 @@
...
 ###
 
+# Number of seconds to wait during sslyze connection check.
+# Not much patience here, and very willing to move on.
+network_timeout = 5
```
konklone: This can be overridden via `--network_timeout`, but unless it becomes used/wanted by anyone, I don't see the need to document it.

scanners/sslyze.py
```diff
@@ -57,42 +73,27 @@ def scan(domain, options):
...
         return None
 
     yield [
         scan_domain,
-        data['protocols']['sslv2'], data['protocols']['sslv3'],
-        data['protocols']['tlsv1.0'], data['protocols']['tlsv1.1'],
-        data['protocols']['tlsv1.2'],
+        data['protocols'].get('sslv2'), data['protocols'].get('sslv3'),
+        data['protocols'].get('tlsv1.0'), data['protocols'].get('tlsv1.1'),
+        data['protocols'].get('tlsv1.2'),
```
konklone: If `sslyze` can't connect to a hostname, all of these fields will be blank.

scanners/sslyze.py
```diff
@@ -224,80 +223,83 @@ def parse_sslyze(raw_json):
...
+
+    return data['certs']
 
 
 # Given the cert sub-obj from the sslyze JSON, use
 # the cryptography module to parse its PEM contents.
 def parse_cert(cert):
     backend = cryptography.hazmat.backends.openssl.backend
-    pem_bytes = cert['as_pem'].encode('utf-8')
+    pem_bytes = cert.public_bytes(Encoding.PEM).decode('ascii').encode('utf-8')
```
konklone: It's possible there's a more efficient way to do this, but I wasn't able to easily figure out one.

scanners/sslyze.py
```diff
@@ -312,12 +314,133 @@ def cert_issuer_name(parsed):
...
+    # Queue them all up
+    queue(Sslv20ScanCommand())
+    queue(Sslv30ScanCommand())
+    queue(Tlsv10ScanCommand())
+    queue(Tlsv11ScanCommand())
+    queue(Tlsv12ScanCommand())
+
+    # Default to cert info on.
+    if options.get("sslyze-no-certs", False) is False:
+        queue(CertificateInfoScanCommand())
```
konklone: The `--sslyze-no-certs` option is similarly undocumented - I used it for debugging, and may again, but I don't expect myself or anyone else to use it, and it may get yanked down the line.

PR #155, opened 2017-11-05
Scanning pipeline refactor, and a Lambda-based pipeline

konklone:
This is a very large PR, as my attempt to introduce a Lambda-based pipeline also led me to a long-overdue refactor of the overall scanning pipeline. 

The Lambda pipeline still relies on one local worker thread per Lambda worker, so there is some local bottlenecking happening. However, local worker threads that are waiting around for Lambda responses are doing much less work than worker threads that are running scans, and so it should be possible to run many more workers when using the Lambda pipeline. (I'm still evaluating the real effects when running large scans.) Local systems may need to expand their system's allowed max number of processes in order to take full advantage.

Future expansions to this pipeline could include a cloud-based queueing system, so that local execution is single-threaded and all parallel activity happens in the cloud. For now, the main benefits of the Lambda pipeline are realized in expanding the number of worker thread substantially, and in eliminating the effects of memory leaks during long-running scans (such as SSLyze's memory leaks when running in synchronous mode).

Some of the general changes:

* All caching of scan data is now performed centrally -- scanners don't need to manage their own caching logic.
* While data is always stored in the `cache/` directory after a scan, the use of cached data to skip the execution of a scan is now **opt-in** instead of **opt-out**. The `--force` flag has been replaced with a `--cache` flag, which enables the use of cached data to skip scans. Use of cached data to skip scans is almost always used only in debugging and development, not in production, so this felt like it made sense to do.
* Scanners can now define an `init` method that gets run for all scans of that type, and an `init_domain` method that gets run for each scan. Both methods can return dicts that will get merged and passed to the scanner's `scan` method as an `environment` dict. The `init_domain` method can return `False` to declare a domain ineligible for that scan and cause it to be skipped.
* A scanner's `scan` method now returns a dict of data, rather than one or more rows. This dict is what is serialized to JSON and cached for that method, so every scanner is responsible for assembling a dict of data to be cached.
* A scanner must now define a `to_rows` method that, given the dict it returned, will transform that dict into one or more rows for CSV output. This decouples CSV-specific logic from other scanning logic.
* A `--meta` flag that causes the resulting CSVs to append extra fields with information about the scan, such as encountered errors and timing information.
* Extra duration and scan info is stored in `meta.json`, including a breakdown per-scanner.

The Lambda pipeline:

* To execute a scan inside of Lambda instead of locally, add the `--lambda` flag. It's (sort of) that simple!
* When using the Lambda pipeline, a scanner's `scan` method will be run inside Lambda. However, all other methods (`init`, `init_domain`, `to_rows`) will be run locally.
* Lambda functions need to be deployed to a specific AWS account before they can be run. Some build scripts to do so, along with a precompiled virtualenv with all required dependencies, are included. Documentation will be added before merge.
* If the Lambda pipeline is being used, `--meta` will also append a bunch of Lambda-related fields, such as the request ID, logging information, and timing information.
* If `--meta` and `--lambda` are passed, `--lambda-details` will cause the scanner to fetch even more post-scan data from Lambda CloudWatch logs, after waiting 20 seconds to allow logs to appear in CloudWatch. This will make one log request per-scan, so can be quite expensive. However, it's the only way to track how much memory individual scan executions actually used.

TODO:

- [x] Add documentation, including for the Lambda deploy pipeline.
- [x] Update `lambda/deploy` to have a `--create` mode that uses the create-function method the first time.
- [x] Update all scanners to use the new format.
- [x] Remove unused scanners (e.g. `tls`).
- [x] Slight refactor of local/lambda branching execution code to reduce duplicated code between both scan methods.
- [ ] ~Make sure Dockerfile still works.~
- [x] If https://github.com/dhs-ncats/pshtt/pull/136 gets merged and `pshtt` is updated on PyPi, update deps in this branch.


konklone:
cc @h-m-f-t @jsf9k @IanLee1521 @egyptiankarim I just rewrote this PR's title and description, and though there are a few TODOs left before merge, review and comments would be much appreciated.

lambda/deploy
```diff
@@ -0,0 +1,61 @@
...
+
+# echo "Updating Lambda handler function for $FUNCTION_NAME..."
+# aws lambda update-function-configuration \
+#   --function-name $FUNCTION_NAME \
+#   --handler lambda_handler.handler
+
+echo "Updating Lambda code file for $FUNCTION_NAME..."
+aws lambda update-function-code \
+  --function-name $FUNCTION_NAME \
+  --zip-file fileb://./build/$FUNCTION_NAME.zip
```
konklone: Noted in a TODO, but this deploy script needs to have a flag to distinguish between creating a function and updating function code in place. A `--create` flag might be enough.

lambda/lambda_handler.py
```diff
@@ -0,0 +1,50 @@
+import importlib
+import sys
+import logging
+
+from scanners import utils
+
+# Central handler for all Lambda events.
+def handler(event, context):
```
konklone: All Lambda function executions begin here, and are then dispatched to the right code file from here. It's up to the Lambda task builder to make sure that this and the scanner file (and `scanner/utils.py`, and the overall set of deps) are present in the task's environment.

lambda/remote_build.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+# On an Amazon Linux box, to initially set up.
+sudo yum install python36 python36-virtualenv
+sudo yum install sqlite-devel gcc libffi-devel openssl-devel
+sudo yum install git
+git clone https://github.com/18F/domain-scan.git
+git clone https://github.com/dhs-ncats/pshtt
+# May need to switch branches, depending.
```
konklone: The above few lines are really here for documentation of how I set up the Amazon Linux instance post-AMI. The rest is repeatable (and to be run on the EC2 instance). This is obviously begging to become a Docker-based process to build these tasks, which is a TODO, but probably for a future PR.

lambda/remote_build.sh
```diff
@@ -0,0 +1,58 @@
...
+tar -zxvf python3.6-sqlite3-3.6.0.tar.gz
+rm python3.6-sqlite3-3.6.0.tar.gz
+
+# Lambda workaround for cryptography (Lambda doesn't have openssl 1.0.2)
+rm -r cryptography/
+rm -r cryptography-1.9-py3.6.egg-info/
+wget https://github.com/Miserlou/lambda-packages/raw/master/lambda_packages/cryptography/python3.6-cryptography-1.9.tar.gz
+tar -zxvf python3.6-cryptography-1.9.tar.gz
+rm python3.6-cryptography-1.9.tar.gz
+
```
konklone: Some obnoxious tweaks above to modify the SQLite, `cryptography`, and general build process to accommodate how the Amazon Linux environment that Lambda runs in works (it's slightly different even from the vanilla Amazon Linux AMI, though it's based on it).

requirements.txt
```diff
@@ -1,14 +1,25 @@
...
+
+# To support a11y scanner:
 pyyaml
 
-# to support sslyze scanner
+# Ubuntu 14.04 has an old version of six
+six>=1.6
+
+# To support pshtt scanner (and to stay fresh):
+git+git://github.com/dhs-ncats/pshtt.git#egg=pshtt
```
konklone: This pulls `pshtt` from GitHub, at least for now, since PyPi has been lagging behind.

Also, this particular build is making use of an unmerged PR to `pshtt`, in https://github.com/dhs-ncats/pshtt/pull/136, which means that any local execution needs to have `pshtt` installed from that branch, and any Lambda execution needs to have had its deps compiled with `pshtt` on that branch (which the .zip in this PR has been).

Once that PR is merged, and ideally when PyPi has been updated, this can all be updated.

scan
```diff
@@ -2,33 +2,76 @@
...
 domain_suffix = options.get("suffix")
 utils.configure_logging(options)
 utils.mkdir_p(utils.cache_dir())
 utils.mkdir_p(utils.results_dir())
 
-# some metadata about the scan itself
-start_time = utils.utc_timestamp()
+# Default and maximum for local workers (threads) per-scanner.
+default_workers = 10
+global_max_workers = 1000
```
konklone: `global_max_workers` is only used for Lambda connection pooling.

scan
```diff
@@ -2,33 +2,76 @@
...
+
+# Some metadata about the scan itself.
+start_time = utils.local_now()
 start_command = str.join(" ", sys.argv)
 
+# Generate a random UUID for the entire scan.
+scan_uuid = str(uuid.uuid4())
```
konklone: Not currently using this anywhere, but for ease of finding logs in Lambda, every scan gets a `uuid`. This is also made for local scans, and stored in `meta.json`.

scan
```diff
@@ -91,74 +129,357 @@ def run(options=None):
...
-        # A scanner can return multiple rows.
-        try:
-            rows = list(scanner.scan(domain, options))
-        except:
-            logging.warn(utils.format_last_exception())
+    # Experimental!
+    if use_lambda:
+        scan_method = perform_lambda_scan
+    else:
+        scan_method = perform_local_scan
```
konklone: This is where things branch for Lambda v local, by assigning a different method to the method that gets sent to each worker. However, eventually there became a lot of copy/pasted code between those two methods. So a TODO is to branch this deeper in by sending the same method to all workers, and branching within that method.

scan
```diff
@@ -91,74 +129,357 @@ def run(options=None):
...
-
+            workers = int(options.get("workers", default_workers))
+
+        # Enforce a local worker maximum as a safety valve.
+        workers = min(workers, global_max_workers)
+
+        # Scan environment, passed to all scanners (local or cloud).
+        environment = {}
+        environment['scan_method'] = {True: 'lambda', False: 'local'}[use_lambda]
+        environment['scan_uuid'] = scan_uuid
```
konklone: The `scan_method` and `scan_uuid` are sent into all scan methods, including Lambda executions. The `scan_method` should be referenced sparingly by a scanner, so as to keep the code agnostic, but for example this is used inside the `sslyze` scanner to default to synchronous execution in Lambda (no multiprocessing) and parallel execution locally (to avoid long-running memory leaks) without forcing the user to keep track of those things.

scanners/noop.py
```diff
@@ -0,0 +1,59 @@
+import logging
+
+###
+# Testing scan function. Does nothing time consuming or destructive,
+# but exercises many of the main hooks of domain-scan.
```
konklone: This `noop` scanner (no operation) is here as an example of how scanners work, and an easy scanner for testing that doesn't make network requests and doesn't risk DoS-ing anyone. It doesn't do anything, but it models how to use the various scanner functions.

scanners/pshtt.py
```diff
@@ -1,82 +1,85 @@
...
-###
-# == pshtt ==
-#
-# Inspect a site's TLS configuration using DHS NCATS' pshtt tool.
-#
-###
+import re
+from pshtt import pshtt
 
-command = os.environ.get("PSHTT_PATH", "pshtt")
```
konklone: In addition to porting the pshtt scanner to work on Lambda, I also ported it to use the Python API rather than shelling out to `pshtt`. We don't reference a `command` or environment variables for anything.

scanners/pshtt.py
```diff
@@ -101,3 +104,7 @@ def scan(domain, options):
     "Base Domain HSTS Preloaded", "Domain Supports HTTPS",
     "Domain Enforces HTTPS", "Domain Uses Strong HSTS", "Unknown Error",
 ]
+
+
+def format_domain(domain):
+    return re.sub("^(https?://)?(www\.)?", "", domain)
```
konklone: This is in place of the `format_domains` function pshtt uses internally, but which only gets run during CLI use of pshtt. It'd be nice of `pshtt` moved this function deeper inside, so that it got run on everything passed to `inspect_domains`.

scanners/sslyze.py
```diff
@@ -1,94 +1,91 @@
...
+#
+# If data exists for a domain from `pshtt`, will check results
+# and only process domains with valid HTTPS, or broken chains.
+#
+# Supported options:
+#
+# --sslyze-serial - If set, will use a synchronous (single-threaded
+#   in-process) scanner. Defaults to true when local, false in cloud.
+# --sslyze-certs - If set, will use the CertificateInfoScanner and
+#   return certificate info. Defaults to true.
```
konklone: `--sslyze-certs` defaults to True, but can be set to False by passing `--sslyze-certs=false` to disable the use of the certificate info scanner.

scanners/sslyze.py
```diff
@@ -345,16 +351,17 @@ def init_sslyze(hostname, options, sync=False):
         return None, None
 
     if sync:
-        scanner = SynchronousScanner()
+        scanner = SynchronousScanner(network_timeout=network_timeout)
     else:
-        scanner = ConcurrentScanner()
+        scanner = ConcurrentScanner(network_timeout=network_timeout)
```
konklone: I passed the `network_timeout` to both scanners. That said, I didn't notice any impact on improving scan times. :/

scanners/sslyze.py
```diff
@@ -367,10 +374,17 @@ def scan_serial(scanner, server_info, options):
...
+    if options.get("sslyze-certs", True) is True:
+
+        try:
+            logging.debug("\t\tCertificate information scan.")
+            certs = scanner.run_scan_command(server_info, CertificateInfoScanCommand())
+        # Let generic exceptions bubble up.
+        except idna.core.InvalidCodepoint:
+            logging.warn(utils.format_last_exception())
+            data['errors'].append("Invalid certificate/OCSP for this domain.")
+            certs = None
```
konklone: This error occurs on `safetyact.gov`.

konklone:
All right, I completed all the tasks in the TODO list - I removed unused scanners, converted the rest to the new format, added docs for the Lambda work (as well as the general changes of adding `--meta` and moving `--force` to `--cache`), and did some needed refactoring of the `scan` script to deduplicate code between both kinds of scan execution.

The `docker-compose up` command now crashes with this:

```
Creating network "scan_default" with the default driver
Recreating scan_scan_1 ... 
Recreating scan_scan_1 ... done
Attaching to scan_scan_1
scan_1  |   File "./scan", line 212
scan_1  |     environment = {**environment, **init}
scan_1  |                     ^
scan_1  | SyntaxError: invalid syntax
scan_scan_1 exited with code 1
```

It appears to be dying on the [new Python unpacking syntax](https://www.python.org/dev/peps/pep-0448/) added in Python 3.5. The Dockerfile appears to use Python 3.6, so I believe the issue is specific to Docker Compose. However, that's the only approach we've documented so far. 

I don't want that to hold up this PR, especially given how much of a mega-PR this is -- the project needs to move forward. So I'll merge and plan to fix Docker issues separately.

@jsf9k if you have any interest in documenting your approach to using docker with this tool in a PR, that'd be very welcome!

PR #156, opened 2017-11-05
docker-compose and Dockerfile

jsf9k:
* `docker-compose.yml`
  * The command `docker-compose build scan` will now build the Docker image.
  * `domain-scan` can be run via a command like `docker-compose run --rm scan scanme.csv --scan=sslyze --serial --debug --force`.
  * No longer fails due to missing `env` directory
* `Dockerfile`
  *  Combined some repetitive lines so that there are fewer layers in the Docker image.  This results in slightly faster builds, slightly smaller images, and most importantly a `Dockerfile` that is a little easier to understand.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,6 @@ MAINTAINER V. David Zvenyach <vladlen.zvenyach@gsa.gov>
 ###
 # Dependencies
 ###
-
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -48,34 +47,22 @@ RUN \
...
-
-RUN apt-get install \
-      -qq \
-      --yes \
-      --no-install-recommends \
-      --no-install-suggests \
+      libncursesw5-dev \
+      # Not sure what these dependencies are for
       nodejs \
       npm
```
konklone: The `nodejs` and `npm` dependencies are for the Node-based scanners in this project (which use `phantomas` as a headless browser). Though I am definitely hesitant to depend on `apt` for that, but that's for another PR.

docker-compose.yml
```diff
@@ -1,5 +1,14 @@
...
+
+services:
+  scan:   
+    build: .
+    volumes:
+      - .:/home/scanner
+    # Some internet providers intercept NXDOMAIN and show a search
+    # page instead.  Google's DNS does not, so I prefer to use it.
+    dns:
+      - 8.8.8.8
```
konklone: I wonder if we could provide DNS server config as a `pshtt` feature flag and handle it in app-layer code...

PR #158, opened 2017-11-05
Eliminate sslyze deadlocks using forkserver

### LINKED ISSUE #157: sslyze calls deadlocking due to combination of threads/processes/logging
There are no more `defunct` processes after #151, and our bulk scans go for much longer before they become an issue, but eventually they do just get stuck. Or as [this says](https://github.com/Birne94/python-atfork-lock-release):

> If a process is forked while an io lock is being held, the child process will deadlock on the next call to `flush`.

After looking at the stuck processes' trace with `gdb`, I'm convinced I'm facing the same issue described here:
https://stackoverflow.com/questions/39884898/large-amount-of-multiprocessing-process-causing-deadlock

And that this bug, opened in 2009 and still quite actively discussed in October 2017, is the cause:
https://bugs.python.org/issue6721

The folks on that bug thread seem to be reaching process at a fix that is specific to logging calls and buffered IO, which I suspect would be enough to fix our case. There's also some related discussion [on this bug](https://bugs.python.org/issue14976#msg300532), with Guido indicating he believes something should be done for the [GC interrupt case](https://codewithoutrules.com/2017/08/16/concurrency-python/).

There are a few ways to work around this I can think of:

* Use sslyze's `SynchronousScanner` instead of the `ConcurrentScanner`. However, this is both much slower and results in a distinct memory leak, as noted in #151.
* Don't ever write to stdout from the child processes. However, this exposes some of my not-totally-complete understanding here -- I am not sure whether it's my own logging calls or something inside `sslyze` that is the problem. Workers at the `domain-scan` level are done as threads via a `ThreadPoolExecutor`, whereas I believe it's SSLyze's `ConcurrentScanner` that forks off processes. So I may have limited control here. The only reference to using stdout in SSLyze's core is [this emergency shutdown message](https://github.com/nabla-c0d3/sslyze/blob/f47806329e5150b5bb67540eabe3ac62781a24c0/sslyze/__main__.py#L28), so I am not sure where in SSLyze this might be happening.
* [This Python module](https://github.com/Birne94/python-atfork-lock-release) that lets you register an after-fork hook that clears up any held locks the child copied over. It's not on PyPi and would need to be installed from the repo (`pip` supports github repo syntax).

Right now, I'm leaning toward the 3rd option, the Python module. I'll try it out and see how it goes.

konklone:
This sets a `domain-scan`-wide use of `forkserver` for any library-initiated use of the Python `multiprocessing` module to fork processes. This creates a sort of master vanilla process that subsequent process forking will use to generate child processes, rather than the calling process. This should resolve #157.

This should work well for us. [The docs for this approach](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) say "set_start_method() should not be used more than once in the program." It also provides `get_context()` as an alternative, but that has to happen in library-level code, where as set_start_method can be used in app-layer code. The docs also say "A library which wants to use a particular start method should probably use get_context() to avoid interfering with the choice of the library user."

Some tradeoffs of this approach:

* If dependencies use `multiprocessing`, and they set their own start method without properly using `get_context()`, that could interfere with this one.
* The processes are a little trickier to understand when viewed in `ps`. There's the main called process, then a near-empty forkserver master process, and then also a child of that forkserver process which appears to be the template from which children are spawned. Killing the main called process does not automatically lead to the death of all generated children, and so multiple `kill` or `killall` statements may be necessary if the processes have to be cleaned up.

Fixes #157. (I think.) Held open until a full bulk scan is completed using this branch, and no deadlocks are observed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,6 @@ pyyaml
 # to support sslyze scanner
 sslyze
 cryptography
-timeout-decorator
 
 # to support censys gatherer
 censys
--- a/scan
+++ b/scan
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import multiprocessing
 import os
 import sys
 import glob
@@ -175,4 +176,10 @@ def domains_from(arg):
 
 
 if __name__ == '__main__':
+
+    # Makes it safer for multi-threaded executions of domain-scan
+    # to make use of the multiprocessing module without causing
+    # deadlocks.
+    multiprocessing.set_start_method('forkserver')
+
     run(options)


konklone:
No deadlocks on a full scan -- and it went very quickly, too! :tada:

konklone:
Hmm. Well, this PR definitely moved things forward and made things much faster and more reliable, but this did just pop up at the very end of the logs:

```
/home/ubuntu/.pyenv/versions/3.6.0/lib/python3.6/multiprocessing/semaphore_tracker.py:129: UserWarning: semaphore_tracker: There appear to be 240 leaked semaphores to clean up at shutdown
```

And it looks like there did turn out to be a couple stuck processes after all, though oddly they didn't seem to hold up the actual scans from finishing. Clearly there's some more work to be done here. :/

gbinal:
Thanks for all of your great work on refactoring this, @konklone.   You're the best.

PR #159, opened 2017-11-18
Adding STARTTLS-related fields to the domain-scan trustymail output

jsf9k:
* Added STARTTLS-related fields to [dhs-ncats/trustymail](https://github.com/dhs-ncats/trustymail) output
* Forcing the use of Google DNS with [dhs-ncats/trustymail](https://github.com/dhs-ncats/trustymail)
* Added the "Errors" field from the [dhs-ncats/trustymail](https://github.com/dhs-ncats/trustymail) output.  This information is useful to have when reconstructing why [dhs-ncats/trustymail](https://github.com/dhs-ncats/trustymail) reached the conclusions it did.  **Note that this also makes this pull request dependent on [pull request 25 from dhs-ncats/trustymail](https://github.com/dhs-ncats/trustymail/pull/25), since it relies on dhs-ncats/trustymail@6d750b77326844850ac01677fa3a66e8d604edf4.**

### DIFF
--- a/scanners/trustymail.py
+++ b/scanners/trustymail.py
@@ -49,7 +49,7 @@ def scan(domain, options):
         data = json.loads(raw)
         utils.write(utils.json_for(data), utils.cache_path(domain, "trustymail"))
 
-    # trustymail scanner follows pshtt in  using JSON arrays, even for single items
+    # trustymail scanner follows pshtt in using JSON arrays, even for single items
     data = data[0]
 
     row = []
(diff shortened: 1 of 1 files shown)


konklone:
I am going to merge this in, so as to avoid merge conflicts in some work I'm about to do on the trustymail scanner to adapt it for the new pipeline in #155 (which is getting merged right after that work is done).

I can use the pshtt branch locally for now, and in the worst case, can revert the changes post-adapting if the pshtt PR gets rejected or changed.

konklone:
(Also, as a side note - I'd like to use the `trustymail` Python API instead of the CLI, as I did with `pshtt`, so that's also something that will change soon.)

PR #175, opened 2017-11-27
The --dns-hostnames option in trustymail was renamed to --dns

jsf9k:
It is used in `domain-scan`, so it also needs to be changed here.


konklone:
Thanks!

PR #176, opened 2017-11-29
Renamed error trustymail output field

jsf9k:
The `trustymail` output field called "Error" was renamed to "Debug".  This requires a change in `domain-scan` as well.

This pull request should be merged once dhs-ncats/trustymail#35 is merged.


konklone:
Thanks, @jsf9k!

Issue #163, opened 2017-11-20
Allow mixing of Lambda and non-Lambda scanners


konklone:
Not all scanners are designed for Lambda use, but it's nice to be able to employ them for the ones that do. If a scanner doesn't support Lambda, it'd be great to just assume local scans for that scanner.

Perhaps scanners can opt-in or -out of registering as Lambda-compatible, so that the orchestrator can override the CLI flag for scanners which aren't registered that way.

brendansudol:
close-able now?

PR #177, opened 2017-11-30
support mixing lambda and non lambda scanners

brendansudol:
This addresses https://github.com/18F/domain-scan/issues/163, allowing one to use `--lambda` for scanners set up for it, and local otherwise.

It adds a `lambda_supported` scanner whitelist which drives a per-scanner `use_lambda` flag.

### DIFF
--- a/scan
+++ b/scan
@@ -22,6 +22,7 @@ from scanners import utils
 # basic setup - logs, output dirs
 options = utils.options()
 domain_suffix = options.get("suffix")
+lambda_mode = options.get("lambda", False)
 utils.configure_logging(options)
 utils.mkdir_p(utils.cache_dir())
 utils.mkdir_p(utils.results_dir())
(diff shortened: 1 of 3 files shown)


scan
```diff
@@ -37,8 +37,13 @@ start_command = str.join(" ", sys.argv)
 # Generate a random UUID for the entire scan.
 scan_uuid = str(uuid.uuid4())
 
+# Whitelist of scanners that can use lambda
+# (facilitates running a mix of lambda / non-lambda scanners together)
+lambda_supported = ["pshtt", "sslyze"]
+lambda_mode = options.get("lambda", False)
```
konklone: Instead of hardcoding the list here, could we add this as an attribute on the `pshtt` and `sslyze` scanner modules? Then any scanner could opt-in to advertising its lambda support without code changes in the central orchestration script.

brendansudol: sure, we can do it that way -- will update

brendansudol: updated!

konklone:
Looks great, thank you!

PR #178, opened 2017-12-02
Renamed "Debug" to "Debug Info"

jsf9k:
The "Debug" output column has been renamed "Debug Info" in trustymail, so it needs to be renamed here too.


konklone:
Thanks for keeping this in sync! (And feedback welcome on superior ways to deal with this issue.)

PR #182, opened 2017-12-19
Fixed bug that made Lambda zips twice as large as they needed to be

jsf9k:
Because the environment zip files were being left in the build directory, the resulting `test_<function>.zip` files were about twice as big as they needed to be.  I added a line to delete the environment zip files after they are extracted.


konklone:
Nice, great catch!

PR #183, opened 2017-12-21
Added AWS Lambda support for trustymail

jsf9k:
**Please merge #182 before this pull request, since this pull request includes the change in that one.**

Note that a side-effect of this pull request is that `domain-scan` now uses the `trustymail` API instead of calling the `trustymail` executable.  That's a plus.

I also made some minor improvements to the `lambda/remote_build.sh` script.

### DIFF
--- a/lambda/deploy
+++ b/lambda/deploy
@@ -47,10 +47,12 @@ if [ -f ../envs/$SCANNER_NAME.zip ]; then
   echo "Incorporating custom-built env for $SCANNER_NAME..."
   cp ../envs/$SCANNER_NAME.zip .
   unzip -q $SCANNER_NAME.zip
+  rm $SCANNER_NAME.zip
 else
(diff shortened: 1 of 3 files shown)


konklone:
This is awesome - thank you. 

This doesn't affect whether it's merge-worthy, but have you found that you can actually effectively run trustymail inside Lambda? I thought Lambda blocked common mail ports to prevent spam. :/

bknowles:
Wouldn't using AWS for this kind of scanning be a violation of their Acceptable Use Policy, as laid out at [https://aws.amazon.com/aup/](https://aws.amazon.com/aup/), specifically the following section:

> Unauthorized Access. Accessing or using any System without permission, **_including attempting to probe, scan, or test the vulnerability of a System_** or to breach any security or authentication measures used by a System.

PR #184, opened 2017-12-21
sslyze should support ports other than 443

jsf9k:
~~**Please merge #182 and #183 before this pull request, since this pull request includes the changes in those ones.**~~

### DIFF
--- a/scan
+++ b/scan
@@ -315,7 +315,8 @@ def perform_scan(params):
         scan_environment = {**environment, **scan_environment}
 
         # If --cache is on, read from this. Always write to it.
-        domain_cache = utils.cache_path(domain, handles[scanner]['name'], ext="json")
+        hostname, port, display_name = parse_domain(domain)
+        domain_cache = utils.cache_path(display_name, handles[scanner]['name'], ext="json")
 
         if (options.get("cache")) and (os.path.exists(domain_cache)):
             logging.warn("\tUsing cached scan response.")
@@ -445,10 +446,13 @@ def write_rows(rows, headers, domain, scanner, csv_writer, meta=None):
         empty_row = [None] * len(scanner.headers)
         rows = [empty_row]
 
-    # Always output Domain and Base Domain.
+    # Always output Domain and Base Domain.  Note that the domain may
+    # contain the port as well.
+    hostname, port, display_name = parse_domain(domain)
+
     standard_prefix = [
-        domain,
-        utils.base_domain_for(domain)
+        display_name,
+        utils.base_domain_for(hostname)
     ]
 
     # If requested, add local and Lambda scan data.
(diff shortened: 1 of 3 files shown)


konklone:
This one touches domain-scan's core code, and I think I'd like to request some changes before merge. Would you be able to rebase this branch so that the changes are just showing those specific to this PR?

jsf9k:
No problem.  Rebased and force pushed.

scan
```diff
@@ -593,5 +597,38 @@ def fetch_lambda_details(dict_row):
...
     return lambda_fields
 
+def parse_domain(domain):
+    """Returns a tuple containing the hostname, port, and display name
+    for the domain.
+
+    The port is assumed to be 443 if it is not present.  The display
+    name is simply the hostname followed by a colon and the port,
+    unless the port is 443.  In that case the colon and port are
+    omitted.
```
konklone: I believe Windows doesn't allow colons in filenames - it looks like this would result in `example.gov:8443.json` files when non-standard ports are used, if I'm reading it right.

jsf9k: You're right.  I try to forget about Windows as much as possible. :penguin: 

We need a character that will not show up in a valid hostname.  Would a space work?

jsf9k:
@konklone I'd actually like to be able to specify the ports.  Having run `trustymail` beforehand, I will know the hostnames and ports that support STARTTLS, and those are the only ones I will want to run `sslyze` against.

> @jsf9k Is it the case that you're essentially using a separate mode for sslyze here? Like, if you were to pass an --sslyze-email flag that would have sslyze do multiple scans to the standard SMTP ports and return a row for each one, would that work?

scan
konklone: A dot seems like it would work all right. (But let's arrive at consensus on an overall strategy first.)

konklone:
> @konklone I'd actually like to be able to specify the ports. Having run trustymail beforehand, I will know the hostnames and ports that support STARTTLS, and those are the only ones I will want to run sslyze against.

OK, this sounds a lot like how we structure TLS scanning, where `sslyze` will rely on previous `pshtt` data to determine which endpoint to scan (the canonical one) or whether it should just skip the scan (if no HTTPS endpoints are supported).

So, `sslyze` could have a default set of SMTP ports it uses for an `--sslyze-email` mode, but it could look to see if there exists any previous `trustymail` scan data and automatically choose only the ports that `trustymail` discovered support STARTTLS.

This is meant to be one of the other advantages of this multi-scanner orchestration, is being able to pass state on from one scanner to the next to make the overall scan process more efficient. We're able to skip tens of thousands of `sslyze` scan attempts on endpoints that `pshtt` was able to tell us don't exist.

Would you be up for trying that approach here? I think it might simplify the overall code, and at least contain it to the `sslyze`/`trustymail` path. I'm also happy to take a pass at it when I have a chance if it's not clear what I mean.

jsf9k:
Sure, I can try that approach.

jsf9k:
@konklone, I reworked this so that it does what I need without touching `scan`.  Let me know what you think.  I'm also happy to rebase and force push to get rid of 577e0d6 and ec08598 if you want.  (Those two commits just cancel each other out anyway.)  Just let me know.

There are two things that I don't like about this code as it stands.  First, you can't pass in a domain like `mail.example.com:25` for one-off.  

Second, if you have two domains `foo.com` and `bar.com` that both use `mail.example.com:25` for their mail server, then that mail server will be tested twice.  The cached results from testing `foo.com` will include the SMTP STARTTLS results for `mail.example.com:25`, but when asked to test `bar.com` `sslyze`/`domain-scan` will have no idea to look for them there.

jsf9k:
@konklone, just checking in to see if you want me to do anything else to get this pull request approved.

konklone:
> First, you can't pass in a domain like mail.example.com:25 for one-off scans.

This is a good point. The tool was formed around the idea that we'd be scanning behavior on a "domain", rather than a specific port, but it's clearly beneficial to be able to pass in port information along with a domain. Even for existing HTTPS/HSTS behavior measurement, or web performance evaluation via headless browsers, we should want the capability to measure differing behavior at `:443` and `:8443` (or other ports). It's rarer for web than for email, but it does happen.

I think that's probably best handled in a separate PR, in a way that deliberately does touch the main `scan` executable and affects what data gets passed into the scanner, as its own feature request.

> Second, if you have two domains foo.com and bar.com that both use mail.example.com:25 for their mail server, then that mail server will be tested twice. The cached results from testing foo.com will include the SMTP STARTTLS results for mail.example.com:25, but when asked to test bar.com sslyze/domain-scan will have no idea to look for them there.

This is true, though we could implement such caching in the `sslyze` scanner if we wanted to. Scanners don't have to exclusively use `cache/sslyze/[domain].json`, they can also read and write to arbitrary locations in `cache/`. So you could also have `cache/sslyze/starttls/mail.example.com.json` files that contain results specific to that mailserver, that an `sslyze` scan for either `foo.com` or `bar.com` can reference when it observes a mailserver record of `mail.example.com`.

Also better in another PR, I think, so as not to block this one.

PR #186, opened 2018-01-03
Using BigQuery to get results from Censys.io

konklone:
This removes all the code related to the Censys.io REST API (which we never use and which I can't really recommend for anything significant, since there's an upper limit on how many results can ever be fetched that way), and all code related to the Censys.io Export API (which is now defunct).

The `censys` gatherer in this PR talks to the Google BigQuery API, and exports hostnames using interactive queries to the `censys-io.certifcates` BigQuery table.

To download data this way, one needs to have been granted access by the Censys.io team, either as an academic researcher account, government researcher account, or as part of a commercial relationship. 

As of right now, this also requires some advance work -- creating a project in Google Cloud Platform, then creating a "service account" with access to create jobs and download data from one's own resources, and then giving that service account identifier to the Censys team for them to grant access to.

Once that's done, the service account JSON data can be set via environment variable, either:

* directly as a string via `BIGQUERY_CREDENTIALS` (e.g. in a PaaS environment), or 
* as a path to a file containing this JSON via `BIGQUERY_CREDENTIALS_PATH` (e.g. in a IaaS environment).

I've also switched this gatherer to make new API requests by default, rather than rely on locally cached data by default, so the `--force` flag has been replaced with a `--cache` flag. This is in line with the changes I made to the `./scan` pipeline in general in #155.

A `--timeout` option can adjust the query timeout from 10 minutes to some other value (in seconds).

Some Censys support pages relevant to this work:

* https://support.censys.io/google-bigquery/bigquery-introduction
* https://support.censys.io/google-bigquery/adding-censys-datasets-to-bigquery
* https://support.censys.io/google-bigquery/certificates-bigquery-dataset

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,32 +188,38 @@ General options:
 * `--ignore-www`: Ignore the `www.` prefixes of hostnames. If `www.staging.example.com` is found, it will be treated as `staging.example.com`.
 * `--debug`: display extra output
 
-### `censys`: the Censys.io API
+### `censys`: Data from Censys.io via Google BigQuery
 
-To configure, set two environment variables from [your Censys account page](https://censys.io/account):
+Gathers hostnames from Censys.io via the Google BigQuery API.
 
-* `CENSYS_UID`: Your Censys API ID.
-* `CENSYS_API_KEY`: Your Censys secret.
+Before using this, you need to:
+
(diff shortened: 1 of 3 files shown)


konklone:
@jsf9k Good eye - just added a commit removing it.

PR #187, opened 2018-01-05
Use with when opening PSL

jsf9k:
A very minor improvement.  Use a `with` clause when opening the PSL so that the file gets closed immediately and does not have to wait to be garbage collected.


konklone:
Thank you, good eye!

PR #189, opened 2018-01-07
Fix for the fact that Python datetimes are formatted differently depending on whether Lambda is used

### LINKED ISSUE #185: Timstamps formatted differently between lambda and local executions
If I run `./scan dhs.gov --scan=sslyze --lambda --debug --workers=400`, I get the following output:
```
Domain,Base Domain,Scanned Hostname,Scanned Port,STARTTLS SMTP,SSLv2,SSLv3,TLSv1.0,TLSv1.1,TLSv1.2,Any Forward Secrecy,All Forward Secrecy,Weakest DH Group Size,Any RC4,All RC4,Any 3DES,Key Type,Key Length,Signature Algorithm,SHA-1 in Served Chain,SHA-1 in Constructed Chain,Not Before,Not After,Highest Served Issuer,Highest Constructed Issuer,Errors
dhs.gov,dhs.gov,www.dhs.gov,443,False,False,False,True,True,True,True,False,256,False,False,False,RSA,2048,sha256,False,False,2017-05-11T00:00:00,2018-08-10T23:59:59,GeoTrust Global CA,GeoTrust Global CA,
```

On the other hand, if I run `./scan dhs.gov --scan=sslyze --debug --workers=400`, I instead get the output:
```
Domain,Base Domain,Scanned Hostname,Scanned Port,STARTTLS SMTP,SSLv2,SSLv3,TLSv1.0,TLSv1.1,TLSv1.2,Any Forward Secrecy,All Forward Secrecy,Weakest DH Group Size,Any RC4,All RC4,Any 3DES,Key Type,Key Length,Signature Algorithm,SHA-1 in Served Chain,SHA-1 in Constructed Chain,Not Before,Not After,Highest Served Issuer,Highest Constructed Issuer,Errors
dhs.gov,dhs.gov,www.dhs.gov,443,False,False,False,True,True,True,True,False,256,False,False,False,RSA,2048,sha256,False,False,2017-05-11 00:00:00,2018-08-10 23:59:59,GeoTrust Global CA,GeoTrust Global CA,
```

Note that **in the lambda results the date and time are separated by a `T` character, while in the local results they are separated by a space character**.  The results should really be identical in the two cases.

jsf9k:
`isoformat()` should use a space as the separator instead of the default `T` character in order to agree with `__str__()`

This resolves #185.

### DIFF
--- a/lambda/lambda_handler.py
+++ b/lambda/lambda_handler.py
@@ -46,7 +46,7 @@ def handler(event, context):
     }
 
     # Serialize and re-parse the JSON, so that we run our own
-    # date transform functions in one place, using the same function
-    # that runs locally.
+    # date transform functions in one place, before Amazon's built-in
+    # JSON serialization prepares the data for transport.
     return utils.from_json(utils.json_for(response))
 
--- a/scan
+++ b/scan
@@ -373,7 +373,15 @@ def perform_scan(params):
 # Let all errors bubble up to perform_scan.
 def perform_local_scan(scanner, domain, handles, environment, options, meta):
     logging.warn("\tExecuting local scan...")
-    return scanner.scan(domain, environment, options)
+    response = scanner.scan(domain, environment, options)
+
+    # Serialize and re-parse data as JSON, to normalize dates
+    # using explicit formatting regardless of local Python environment.
+    #
+    # This is also done for Lambda scans, but performed server-side
+    # by the Lambda handler so that it's done before Amazon's own
+    # JSON serialization is used for data transport to the client.
+    return utils.from_json(utils.json_for(response))
 
 
 ###
--- a/scanners/utils.py
+++ b/scanners/utils.py
@@ -104,7 +104,7 @@ def from_json(string):
 
 def format_datetime(obj):
     if isinstance(obj, datetime.date):
-        return obj.isoformat()
+        return obj.isoformat(' ')
     elif isinstance(obj, str):
         return obj
     else:
--- a/scanners/utils.py
+++ b/scanners/utils.py
@@ -104,7 +104,7 @@ def from_json(string):
 
 def format_datetime(obj):
     if isinstance(obj, datetime.date):
-        return obj.isoformat(' ')
+        return obj.isoformat()
     elif isinstance(obj, str):
         return obj
     else:


bknowles:
With respect, the ISO-8601 format really does use "T" as the separator.  The page at  https://en.wikipedia.org/wiki/ISO_8601 does accurately reflect this fact.

jsf9k:
@bknowles that is true, but the point of this pull request is to get the Lambda-produced results to agree with the non-Lambda-produced results.  I'm not opposed to changing the non-Lambda results to agree with the Lambda results, but that will break a lot of existing code that uses domain-scan.

I'll let @konklone decide.  I'm fine with either solution.

On a side note, Python's choice to have the output of `datetime.__str__()` disagree slightly with that of `datetime.isoformat()` is odd.  The by-the-book ISO format seems like a natural choice for `datetime.__str__()`.

bknowles:
I am almost certain that @konklone was right when he said:

> I'm going to guess it's related to a Python version/library difference between the Lambda environment and your local computer caused by relying on some implicit behavior somewhere instead of us explicitly specifying a timestamp format.

If it were me, I would force ISO 8601 format everywhere and then we could reliably parse on the "T" as the separator, instead of trying to guess what the separator should be, or trying to force the old format.  But maybe that's just me.

konklone:
> If it were me, I would force ISO 8601 format everywhere and then we could reliably parse on the "T" as the separator, instead of trying to guess what the separator should be, or trying to force the old format. But maybe that's just me.

Yeah, I would prefer to be explicit here -- and I would prefer to use the `T` instead of the space, as my experience building API servers/clients that produce/consume timestamps has been that this is what clients expect and that it's closest to the spec.

I can reproduce the behavior (of having a space separator) in my own local environment as well - I'm taking a look at fixing it that way myself.

konklone:
@jsf9k I just added a commit to this PR that alters it to use JSON re-serialization in both code paths, to get explicit date formatting that uses `T` as the separator.

Mind giving it a once-over and verifying it works as expected?

jsf9k:
I tested and this seems to work.  It generates the {{T}} in both cases (lambda and non-lambda).

PR #190, opened 2018-01-08
New trustymail fields

jsf9k:
Updated to include some new fields output by the latest version of [`trustymail`](https://github.com/dhs-ncats/trustymail).


konklone:
:+1: Thank you!

PR #191, opened 2018-02-12
Add check for Content-Security-Policy header

dhilton:
__What is the context of the pull request?__

As a response to the [BrowseAloud hack](http://www.bbc.co.uk/news/technology-43025788) which loaded a JS cryptocurrency mining malware, this commit adds a new scanner to check for the existance of the Content-Security-Policy header which seeks to limit the ability of dangerous content from execution.

This is an initial version that merely checks for the canonical header as currently implemented by common browsers. Future work should look out for common gotchas in wild card settings, etc. Comments / improvements welcome!

__How to test__

1. Check out the branch
2. Run a scan against Mozilla: `./scan mozilla.org --scan=csp`
3. Check that the output csv contains a True flag for mozilla.org `less ./results/csp.csv `
4. Run a fresh scan against gov.uk `./scan gov.uk --scan=csp`
5. Check that the output csv contains a False flag for gov.uk `less ./results/csp.csv ` (correct at PR checktime)

### DIFF
--- a/scanners/csp.py
+++ b/scanners/csp.py
@@ -1,0 +1,71 @@
+import logging
+import requests
+from scanners import utils
+
+###
+# CSP Scanner - check the presence of CSP headers
+#
+
+
+# Set a default number of workers for a particular scan type.
+# Overridden by a --workers flag.
+workers = 2
+
+
+# default to a custom user agent, can be overridden
+user_agent = "github.com/18f/domain-scan, csp.py"
+
+
+def init_domain(domain, environment, options):
+    # If we have data from pshtt, skip if it's not a live domain.
+    if utils.domain_not_live(domain):
+        logging.debug("\tSkipping, domain not reachable during inspection.")
+        return False
+
+    # If we have data from pshtt, skip if it's just a redirector.
+    if utils.domain_is_redirect(domain):
+        logging.debug("\tSkipping, domain seen as just an external redirector during inspection.")
+        return False
+
+    # requests needs a URL, not just a domain.
+    url = None
+    if not (domain.startswith('http://') or domain.startswith('https://')):
+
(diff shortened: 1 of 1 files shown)


scanners/csp.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # requests needs a URL, not just a domain.
+    url = None
+    if not (domain.startswith('http://') or domain.startswith('https://')):
+
+        # If we have data from pshtt, use the canonical endpoint.
+        if utils.domain_canonical(domain):
+            url = utils.domain_canonical(domain)
+
+        # Otherwise, well, ssl should work.
```
konklone: You sure? :) It's usually safer to default to `http://` than `https://`, without information helping you pick which one.

dhilton: On this point I'm assuming that any uk gov.uk domain (my current daytime world) will be over TLS; the protection of data in transit is a core cloud security principle in the UK https://www.ncsc.gov.uk/guidance/cloud-security-principle-1-data-transit-protection but am happy to add a retry for first TLS and then plain text?

konklone: I admire what the NCSC and GDS have been doing to encourage the use of encryption in the UK government, but I highly doubt that all gov.uk subdomains (and other UK government sites) will have HTTPS enabled. But I do suspect that most/all will redirect port 80 to 443. You'll likely get more complete data defaulting to port 80 and following redirects.

But this is also only for the case where no `pshtt` data exists. So you can ignore this issue entirely by including `pshtt` as a scan in your pipeline, for example running `./scan domain.gov.uk --scan=pshtt,third_parties`, which will allow `pshtt`'s calculation of the site's "canonical" URL to determine what URL gets scanned by `phantomas` here.

PR #204, opened 2018-03-13
Latest sslyze drop broke sslyze scanner

jsf9k:
The latest version of sslyze (v1.4.0) has an API change that broke the sslyze scanner.  This commit fixes that.

See these links for details:
* https://github.com/nabla-c0d3/sslyze/releases/tag/1.4.0
* https://gist.github.com/nabla-c0d3/91d6544018e75efe4385b2f4409854ab

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+lambda
+venv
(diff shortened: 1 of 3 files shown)


konklone:
Thank you for keeping on top of this!

PR #205, opened 2018-03-15
Add sslyze minimum version to requirements

jsf9k:
Add sslyze minimum version to `requirements.txt`.


konklone:
Thanks, good addition!

PR #209, opened 2018-03-20
Specifying PSL cache location in trustymail

jsf9k:
Forcing `trustymail` to use the same PSL cache as the rest of `domain-scan` instead of its own.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -95,17 +95,22 @@ RUN echo 'eval "$(pyenv init -)"' >> /etc/profile \
 #     /opt/pyenv/sources/${PYENV_PYTHON_VERSION}-debug/Python-${PYENV_PYTHON_VERSION}/ \
 #     >> etc/gdb/gdbinit
 
+###
+# Update pip and setuptools to the latest versions
+###
+RUN pip install --upgrade pip setuptools
+
+###
+# domain-scan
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -125,15 +130,15 @@ RUN npm install --global phantomas phantomjs-prebuilt \
 ###
 RUN pip install --upgrade \
     git+https://github.com/dhs-ncats/pshtt.git@develop \
-    git+https://github.com/dhs-ncats/trustymail.git@develop
+    git+https://github.com/dhs-ncats/trustymail.git@bugfix/problem_with_psl_caching
```
konklone: It looks like this was merged into `develop` already: https://github.com/dhs-ncats/trustymail/pull/72

Should this line get reverted back to `develop`?

jsf9k: Yes, that wasn't intended to be committed.  Thanks for catching that!

jsf9k: @konklone, it looks like you'll have to un-request changes before I can merge.

konklone:
Ah yeah, I set the `master` branch as protected. If anything comes across as too restrictive, very happy to discuss!

jsf9k:
No, I think it's good to do things this way.

PR #210, opened 2018-03-20
Make flake8 ignore virtualenv directories.

tadhg-ohiggins:
I use pipenv locally, so excluding these would be helpful.


konklone:
:+1:

PR #211, opened 2018-03-21
Move to pytest

tadhg-ohiggins:
Switch over from unittest to pytest.

Also switch to the test setup from http://docs.python-guide.org/en/latest/writing/structure/ (in particular, the use of some path manipulation and the `context.py` file).

### DIFF
--- a/tests/test_process_a11y.py
+++ b/tests/test_process_a11y.py
@@ -1,6 +1,6 @@
 import unittest
 
-from .context import utils  # noqa
+from context import utils  # noqa
 
 from utils.a11y.process_a11y import A11yProcessor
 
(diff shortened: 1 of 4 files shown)


konklone:
:+1: !

tests/test_process_a11y.py
```diff
@@ -1,5 +1,7 @@
 import unittest
 
+from context import utils  # noqa
```
jsf9k: What does "noqa" mean?

konklone: Looks like it asks flake8 to leave the line alone: https://github.com/PyCQA/pycodestyle/issues/476

PR #214, opened 2018-03-22
Use argparse for gather command parameters

tadhg-ohiggins:
The Python standard
Library has many things.
Argparse isn't bad.

Switch the gather command over to using argparse for handling its arguments, and add tests.

My main fear here is that there's some obscure invocation I've broken in making this change, so please let me know if I've missed some corner cases.

### DIFF
--- a/gather
+++ b/gather
@@ -11,7 +11,7 @@ import importlib
 
 from utils import utils
 
-options = utils.options()
+options = utils.options_for_gather()
 utils.configure_logging(options)
 utils.mkdir_p(utils.cache_dir())
 utils.mkdir_p(utils.results_dir())
@@ -48,11 +48,11 @@ def run(options=None):
         os.remove(result)
 
     # Opt in to include parent (second-level) domains.
-    include_parents = options.get("include-parents", False)
+    include_parents = options.get("include_parents", False)
 
     # Opt into stripping www. prefixes from hostnames, effectively
     # collapsing www.[host] and [host] into one record.
-    ignore_www = options.get("ignore-www", False)
+    ignore_www = options.get("ignore_www", False)
 
     # --parents should be a CSV whose first column is parent domains
     # that will act as a whitelist for which subdomains to gather.
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,41 @@
+import sys
+import pytest
+from .context import utils  # noqa
+from utils import utils as subutils
+
+
+@pytest.mark.parametrize("args,expected", [
+    ("gather dap", {"_": ["dap"]}),
+    (
+        "".join([
+            "gather dap --suffix=.gov --dap=",
+            "https://analytics.usa.gov/data/live/sites-extended.csv"]),
+        {
+            "_": ["dap"],
+            "suffix": ".gov",
+            "dap": "https://analytics.usa.gov/data/live/sites-extended.csv",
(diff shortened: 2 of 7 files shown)


gather
```diff
@@ -48,11 +48,11 @@ def run(options=None):
         os.remove(result)
 
     # Opt in to include parent (second-level) domains.
-    include_parents = options.get("include-parents", False)
+    include_parents = options.get("include_parents", False)
 
     # Opt into stripping www. prefixes from hostnames, effectively
     # collapsing www.[host] and [host] into one record.
-    ignore_www = options.get("ignore-www", False)
+    ignore_www = options.get("ignore_www", False)
```
konklone: Are hyphens for arguments no good? This will break some invocations, though it's obviously doable to update/notify downstream users if need be.

tests/test_utils.py
```diff
@@ -0,0 +1,179 @@
...
+            'dap': 'https://analytics.usa.gov/data/live/sites-extended.csv',
+            'private': '/path/to/private-research.csv',
+            'parents': 'https://github.com/GSA/data/raw/master/dotgov-domains/current-federal.csv',
+        }
+    ),
+    (
+        "".join([
+            "./gather dap --suffix=.gov --dap=",
+            "https://analytics.usa.gov/data/live/sites-extended.csv",
+            " --ignore-www",
```
konklone: Maybe I'm not intuiting what the change means -- do `--ignore-www` and `-ignore_www` both work now?

utils/utils.py
```diff
@@ -79,6 +82,103 @@ def options():
...
+
+def options_for_gather():
+    # Parse options for the ``gather`` command.
+    set_services = ("censys")
+    services = [s for s in sys.argv[1].split(",") if s not in set_services]
+    parser = build_gather_options_parser(services)
+    parsed, remaining = parser.parse_known_args()
+    for remainder in remaining:
+        if remainder.startswith("--"):
+            raise argparse.ArgumentTypeError(
```
konklone: I guess I was expecting a little more magic from the options parser, then having to ever check for `.startswith("--")`. Is there something unusual this is trying to do to preserve the `gather` syntax?

gather
tadhg-ohiggins: The hyphens on the command-line still work fine, it's just that argparse will automatically transform them into underscores so that they work as Python object property names. No change on the user side should be necessary.

tests/test_utils.py
tadhg-ohiggins: On the command line only `--ignore-www` will work; `ignore_www` is what the property name will be after argparse processes them.

utils/utils.py
tadhg-ohiggins: This check is probably being overly-cautious, but because we're dynamically creating new options here based on what gatherer is specified (for example, `dap` means that `--dap` is required), we have to account for options that the parser doesn't recognize, and this clause is trying to catch some stray flag that isn't a gatherer that the user might have entered.

It's possible that instead of dynamically turning gatherers into arguments, what we should be doing is scanning the `gatherers` directory for its contents first and constructing arguments out of that. I'm not yet sure about which of those would be better. It wouldn't be that big a change to switch over, though.

konklone: That makes sense. Thanks for the explanation. I'm not sure which is better either, and this works for now either way.

PR #216, opened 2018-03-24
switch to circleci 2.0 format

JasonMWhite:
The CircleCI 1.0 format is being sunsetted in August. This PR updates the CircleCI config file to use the new 2.0 format, and uses a workflow to enable parallel testing in both Python 3.5 and Python 3.6.

For reference: https://circleci.com/blog/sunsetting-1-0/

Please let me know if there's anything else I need to do to begin contributing!


JasonMWhite:
If it's easier to do this in 2 PRs (so as to transition the required status checks), that's easy enough.

PR #217, opened 2018-03-25
Reverting to older sslyze

jsf9k:
The new 1.4 version of sslyze seems to eventually segfault when I scan.  As a result I'm reverting to the latest 1.3 version.  

This issue is impacting the NCATS HTTPS and Trustworthy Email scanning.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,7 +25,7 @@ ipython
 git+git://github.com/dhs-ncats/pshtt.git#egg=pshtt
 
 # sslyze
-sslyze>=1.4.0
+sslyze>=1.3.4,<1.4.0
 cryptography
 
 # a11y
--- a/scanners/sslyze.py
+++ b/scanners/sslyze.py
@@ -14,7 +14,7 @@
 
 import logging
 
-from sslyze.server_connectivity_tester import ServerConnectivityTester, ServerConnectivityError
+import sslyze
 from sslyze.synchronous_scanner import SynchronousScanner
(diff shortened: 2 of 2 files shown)


jsf9k:
See also the related pull request dhs-ncats/pshtt#158.

konklone:
@jsf9k Can you get an isolatable test case for the segfault? We should definitely report that upstream.

For the API change, it looks like this is [documented in the release notes](https://github.com/nabla-c0d3/sslyze/releases/tag/1.4.0) and [they have a migration example](https://gist.github.com/nabla-c0d3/91d6544018e75efe4385b2f4409854ab).

It'd be nice to get 1.4.1 support, since it [has support for detecting deprecated Symantec-chained certificates](https://github.com/nabla-c0d3/sslyze/releases/tag/1.4.1), while also being [smart enough to use a CA blacklist](https://github.com/nabla-c0d3/sslyze/commit/5b7c70fd835db4289356d5365a5f36283d5d709b) (that properly ignores Apple and Google certs, which are not affetced by the distrust).

jsf9k:
@konklone I haven't had a chance to isolate the problem yet.  I've been busy figuring out the issue and getting this weekend's scan results produced.  I'll be looking into it this week.

I saw the migration example regarding the API change.  That's how I knew the change to make in 059625294c0e70a1ee216b4c7cd811ccf4775d96, which I reverted in this pull request.  I'll revert the revert in a separate pull request once I figure out what is going on and we can safely move to `sslyze` 1.4+.

PR #220, opened 2018-03-25
break write_rows out of scan.sh

JasonMWhite:
This PR mainly just forklifts `write_rows` out of `scan` into a new module at `runner/runner.py` (name suggestions welcome). I'd like to see this become the place where most of the code from `scan` lands. As well, this is already showing hints of possible areas for refactoring.

@konklone @tadhg-ohiggins


requirements.txt
```diff
@@ -47,3 +47,4 @@ google-auth-oauthlib
 ############
 
 pytest
+typing-extensions
```
JasonMWhite: Pretty sure this will break the lambda function as-is, since this isn't vendored in yet. Not sure how to do that - any suggestions?

If this is true, it also indicates a missing test. Shouldn't be too hard to write.

runner/runner.py
```diff
@@ -0,0 +1,47 @@
+import typing
+from utils import utils
+from typing_extensions import Protocol
+
+
+class ScannerProtocol(Protocol):
+    headers = []  # type: typing.List[str]
```
JasonMWhite: If you haven't seen this before, this is from [PEP 544](https://www.python.org/dev/peps/pep-0544/). Allows type checkers to check static duck typing. It's standard in Python 3.7, backported to 3.5/3.6 through the `typing-extensions` package.

scan
```diff
@@ -362,7 +361,7 @@ def perform_scan(params):
         # If --meta wasn't requested, throw it all away.
         if not options.get("meta", False): meta = None
 
-        write_rows(rows, handles[scanner]['headers'], domain, scanner, handles[scanner]['writer'], meta=meta)
+        runner.write_rows(rows, domain, utils.base_domain_for(domain), scanner, handles[scanner]['writer'], meta=meta)
```
JasonMWhite: Couple of differences here:
- `write_rows` wasn't using the `headers` argument at all, so I removed it
- `write_rows` was pulling options directly from `argparse`. That's really difficult to test, and makes the function less atomic, so I pulled it out. I imagine we'll see a lot of that as we go.

requirements.txt
konklone: I don't think it would matter, as the Lambda function never relies on any code that make use of these extensions. All that gets uploaded to Lambda are `lambda/handler.py`, `scanners/[scanner].py`, `utils/*.py`, and whatever vendored requirements are needed to support those things. 

Also,I think it would only be the flake8 or testing code path that would validate these things, and nothing that runs in Lambda.

JasonMWhite: When running the code, all of the typing code is effectively a no-op. But we'd still have to be careful, since the `typing_extension` is still being imported at the head of `runner.py`.

If you say so, I'll believe you :)

konklone:
I think the forklifting of, and changes to, `write_rows` are good, especially in making it more atomic. (It's one of the jankier methods.)

Though, the duck-typing code here is on the heavy side. I think capturing typing for key variables -- particularly methods used by individual scanners -- can be good, but I don't think marking out the `meta` arg as `typing.Optional[typing.Mapping[str, typing.Any]]` is as necessary. This is still a small system, and `write_row` is an internal method that user-written scanners won't interact with.

Would you mind splitting the duck-typing code out until later on?

The tests are a help, thank you. In general, I think the best ROI on tests for this project will be comprehensive functional/integration tests to validate intended behavior under a variety of conditions, and unit tests  on critical-path code. The ideal is to be able to refactor internals without rewriting a lot of tests, while still covering the parts of the code that are the most complex or easy to regress.

requirements.txt
konklone: Yeah, `runner/runner.py` won't be imported by a Lambda function as packaged. During Lambda execution, a lot of code still runs locally -- all that gets run in Lambda is a scanner's `scan()` function, and the bare minimum supporting code to make that work. The CSV conversion, meta output, etc., all happens locally even when run in --lambda` mode.

JasonMWhite:
Split out the duck-typing code specifically (the `Protocol`), or the typing code overall? The line you quote isn't related to duck-typing. I'm fine pulling either out at this time, just wanted to introduce what it would look like.

I'm personally rather in favour of documenting for the reader what the expected types are for each argument, as this tends to illustrate rather clearly where methods should probably be cleaned up. (`typing.Optional[typing.Union[str, int, typing.Dict[str, typing.Any]]] anyone? :) )`. Happy to punt on this decision to a later time, though.

I think we are aligned on what to test overall. I like primarily testing public methods, and taking any methods private that are mere internal implementation details. `write_rows` very much falls into the second camp, and I think these tests will eventually become integration tests. One step at a time, though.

konklone:
> Split out the duck-typing code specifically (the Protocol), or the typing code overall? 

Sorry, yes, the typing code overall. If you're okay punting on this for a bit, I think that'll give us time to talk (maybe in Slack? we've got an HTTPS room and can have guests!) about what the ideal state looks like here.

> I think we are aligned on what to test overall. I like primarily testing public methods, and taking any methods private that are mere internal implementation details. write_rows very much falls into the second camp, and I think these tests will eventually become integration tests. One step at a time, though.

:+1: :footprints:

JasonMWhite:
Ready for a second pass.

PR #223, opened 2018-03-27
Removing the explicit pshtt dependency from requirements.txt

jsf9k:
`trustymail` isn't included in this way, and it often breaks things when I'm testing and want to use a different version of `pshtt`.  It will also save some space for those creating Docker containers for scanning that do not require `pshtt`.

I'm also open to removing the explicit `sslyze` dependency, for similar reasons.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,9 +21,6 @@ ipython
 # Scanners
 ############
 
-# pshtt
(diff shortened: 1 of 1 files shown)


jsf9k:
What do you think about removing sslyze too, then?

PR #225, opened 2018-03-27
Use argparse for scan

tadhg-ohiggins:
If a file should fall
In a directory tree
Will anyone know?

Move option handling for `scan` command to argparse, add help info for both `scan` and `gather` commands, refactor to reduce reliance on global variables.

My primary worry with this is that there's some `./scan` edge case, especially around setting a custom cache directory, that I've missed, since there are a lot of small changes in this.

### DIFF
--- a/gatherers/censys.py
+++ b/gatherers/censys.py
@@ -2,6 +2,7 @@ import os
 import json
 import csv
 import logging
+from typing import List
 
 from google.cloud import bigquery
 from google.oauth2 import service_account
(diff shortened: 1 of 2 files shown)


utils/utils.py
```diff
@@ -176,6 +228,136 @@ def options_for_gather():
...
+                "Can't set lambda profile unless lambda flag is set.")
+
+    # We know we want one value, but the ``nargs`` flag means we get a list.
+    should_be_singles = [
+        "lambda_profile",
+        "output",
+        "scan",
+        "suffix",
+        "workers",
+    ]
```
konklone: This is a lot of custom logic controlling how arguments are parsed. I would rather change the flags to make it easier to parse them consistently, then to create a sophisticated parser that molds itself around how they are currently used. 

I understand why the dynamic gatherer names need some special code, but not why the "singles" need their own code. Is there something we should look at changing in the CLI interface to make argument parsing easier?

gather
```diff
@@ -29,22 +24,15 @@ strip_wildcard = re.compile("^(\*.)+")
 strip_redacted = re.compile("^(\?\.)+")
 
 
-def run(options=None):
+def run(options=None, cache_dir="./cache", results_dir="./results"):
```
konklone: It looks like `"./cache"` appears over 20 times in this PR, generally specifying a default value, which seems unwieldy.

It's also not the right level to be specifying values all over the place. The `cache_dir` and `results_dir` must always be subdirectories of the `--output` directory (which defaults to `.`). There's no way to specify separate `cache/` and `results/` locations, and they can always be derived as a direct subdirectory of `--output`.

So given that, what I would expect to see is the default `--output` value specified once, and then cache/results directories dynamically calculated as needed from that value. If there's concern about using global variables, then this value could be stored in a more intentional place and imported (a top-level module where config values/constants are drawn from? some other kind of config file for project values/defaults?).

scan
```diff
@@ -240,7 +250,9 @@ def scan_domains(scanners, domains, options):
         # Kick off workers in parallel. Returns when all are done.
         scan_start_time = utils.local_now()
         with ThreadPoolExecutor(max_workers=workers) as executor:
-            tasks = ((scanner, domain, handles, environment, options) for domain in domains_from(domains))
+            tasks = ((scanner, domain, handles, environment, options)
+                     for domain in domains_from(
+                         domains, domain_suffix=options.get("suffix")))
```
konklone: I don't want to artificially line-wrap lines this way, and I think they're already exempted in flake8 checks. While this might make things easier in certain code editors, this makes the code less readable overall.

scan
```diff
@@ -359,12 +372,16 @@ def perform_scan(params):
...
 
         # If --meta wasn't requested, throw it all away.
-        if not options.get("meta", False): meta = None
+        if not options.get("meta", False):
+            meta = None
 
-        runner.write_rows(rows, domain, utils.base_domain_for(domain), scanner, handles[scanner]['writer'], meta=meta)
+        runner.write_rows(rows, domain,
+                          utils.base_domain_for(domain, cache_dir=cache_dir),
+                          scanner, handles[scanner]['writer'], meta=meta)
```
konklone: If this one's going to be broken up over multiple lines (and I can see the value from a readability perspective), it's more readable with indentation like this:

```python
runner.write_rows(
    rows, domain,
    utils.base_domain_for(domain, cache_dir=cache_dir),
    scanner, handles[scanner]['writer'], meta=meta
)
```

I could see breaking it up into more lines than that, but the important thing is not to base the indentation of the arguments on the length of the name of the method call being made.

scanners/a11y.py
```diff
@@ -17,6 +17,7 @@
 def init(environment, options):
     global redirects
     global config
+    cache_dir = options.get("_", {}).get("cache_dir", "./cache")
```
konklone: This is too much logic for individual scanners to need to know about in order to look up stuff in the cache directory. The domain-scan system should provide this as an easy utility method, global variable, or importable variable.

tests/test_utils.py
```diff
@@ -161,19 +253,113 @@ def test_options_for_gather_missing_arg_parameter(monkeypatch, args):
...
+            "scan": "a11y",
+            "output": "..",
+            **scan_default_false_values,
+        }
+    ),
+])
+def test_options_for_scan_check_for_single_args(monkeypatch, command, expected):
+    monkeypatch.setattr(sys, "argv", command.split(" "))
+    result = subutils.options_for_scan()
+    assert result == expected
```
konklone: I definitely want to get more tests in here, and a few simple tests making sure our command line parsing works properly makes sense. But this seems like a lot of testing code for what in most tools is a routine parsing process that uses some auto-inferrable template for specifying options. I think this is another manifestation of my comment above, that the argument parsing code has gotten much more complex than expected.

scan
tadhg-ohiggins: I don't have a strong preference between these two approaches and am happy to change it.

utils/utils.py
tadhg-ohiggins: I could be missing something, but I don't think that argparse lets us get away from this particular quirk. We use `nargs` to tell it that we expect some number of arguments (usually 1) to that flag, and it returns a list from any use of `nargs` even if the value is `1`. The other option here was changing code further down the chain to expect lists rather than single values, but I felt it was better to address it in a single spot.

tadhg-ohiggins: I'm happy to simplify the code by changing the CLI as it is; I had been trying to avoid that while working on this PR.

gather
tadhg-ohiggins: It is messy to have it in a lot of places, but I wanted to use the explicit & repetitive approach first to replace the use of global variables that end up being accessed from relatively deep in the execution path, and make sure that our tests cover their being passed along correctly, before moving to cleaner options.

scan
tadhg-ohiggins: I'm pretty sure I wrapped that because it's over 100 characters wide, which I think is what's set in our flake8 config.
I'm not sure I agree on readability, but I'm very used to working on files limited to being 80 characters wide.

tests/test_utils.py
tadhg-ohiggins: These tests are in place to cover the functionality that was present before I changed the argument-parsing code; I tried not to add any new functionality, just to make tests that addressed the various edge cases I found in `scan` and the scanners.

gather
tadhg-ohiggins: A cleaner approach might look like something I was working on for gatherers and that I just uploaded as https://github.com/18F/domain-scan/pull/229

PR #226, opened 2018-03-27
Move process_a11y script to its own directory

tadhg-ohiggins:
Preliminary step to make `process_a11y` work out of a separate directory. The next step is to move it out of the project entirely.


konklone:
Pulling in @micahsaul to review this one. Should be straightforward for this PR since it's just directory changes. But I think the plan next is to move this out of the project and into `pulse/`, and this would also be a good time for thoughts on whether that's an okay plan, or what alternatives exist.

PR #227, opened 2018-03-29
Censys tests

tadhg-ohiggins:
Whitespace is tricky
So are variable types.
Try some mypy help.

Add some basic tests for the Censys gatherer, do some minor refactoring, and add some small type hints along the way.

### DIFF
--- a/gatherers/censys.py
+++ b/gatherers/censys.py
@@ -130,27 +130,28 @@ def gather(suffixes, options, extra={}):
 
 def query_for(suffixes):
 
-    select = """
-    parsed.subject.common_name,
-    parsed.extensions.subject_alt_name.dns_names
-    """
+    select = "\n".join([
+        "    parsed.subject.common_name,",
+        "    parsed.extensions.subject_alt_name.dns_names",
+    ])
 
-    from_clause = """
-    `censys-io.certificates_public.certificates`,
-    UNNEST(parsed.subject.common_name) AS common_names,
-    UNNEST(parsed.extensions.subject_alt_name.dns_names) AS sans
-    """
+    from_clause = "\n".join([
+    "    `censys-io.certificates_public.certificates`,",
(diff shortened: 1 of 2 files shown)


gatherers/censys.py
```diff
@@ -2,6 +2,7 @@
 import json
 import csv
 import logging
+from typing import List
```
konklone: Will this lock us to Python 3.7+ as written?

gatherers/censys.py
```diff
@@ -127,45 +128,61 @@ def gather(suffixes, options, extra={}):
...
-    from_clause = """
-    `censys-io.certificates_public.certificates`,
-    UNNEST(parsed.subject.common_name) AS common_names,
-    UNNEST(parsed.extensions.subject_alt_name.dns_names) AS sans
-    """
+    from_clause = "\n".join([
+        "    `censys-io.certificates_public.certificates`,",
+        "    UNNEST(parsed.subject.common_name) AS common_names,",
+        "    UNNEST(parsed.extensions.subject_alt_name.dns_names) AS sans",
+    ])
```
konklone: I don't have strong feelings either way, but what makes `"\n".join([ ... ])` better than using `"""..."""` syntax for long multi-line strings?

gatherers/censys.py
tadhg-ohiggins: No, typing supported this some time back. It definitely works in my local 3.6 environment, and I'm sure it works in 3.5 as well.

gatherers/censys.py
tadhg-ohiggins: It's much easier to see whitespace differences. With the multi-line strings you have to conform to the indentation of the surrounding block.
Using the join approach makes it very clear when you want to add space on purpose versus when you're just doing that to satisfy the parser.

konklone:
Thanks, that addressed my questions!

PR #228, opened 2018-03-29
Test rdns

tadhg-ohiggins:
Very basic tests for the rdns.py gatherer, with some minor refactoring thrown in.


gatherers/rdns.py
```diff
@@ -1,6 +1,7 @@
 import json
 import logging
 import re
+from typing import Generator, List, Pattern
```
konklone: Will this lock the project to Python 3.7+ as written?

tadhg-ohiggins: It definitely works in 3.6 and I'm pretty sure 3.5 as well, if not earlier versions.

tadhg-ohiggins:
I didn't test it empirically and a Reverse DNS data file would be quite useful, thanks.

konklone:
> I didn't test it empirically and a Reverse DNS data file would be quite useful, thanks.

Cool -- remind me when we check in next, when I will also be on a GSA network and can get to the last example I used. (Which is what produced the snapshot CSV at https://github.com/GSA/data/blob/master/dotgov-websites/rdns-federal-snapshot.csv.)

PR #229, opened 2018-03-29
Sketch of an abstract base class approach for gatherers

tadhg-ohiggins:
Sometimes the parent
Should act as the example
The child implements.

The gatherers don't need this kind of approach as much as the scanners do, I think, but trying it with the gatherers first should get the idea across of how we could tackle moving to a more object-oriented approach, with the advantage that the gatherer and scanner objects would be able to store e.g. options and suffixes without having to either pass them around or use global variables.


tadhg-ohiggins:
This is a speculative PR, which is why it renames `gather` to `gatherabc` and adds a `gatherersabc` directory. If we decide to go in this direction those would replace the current `gather` and `gatherers` directories, but this was easier to get going quickly.

jsf9k:
I wholeheartedly support a more object-oriented approach, particularly for the scanners.  I'm also digging the haikus. :+1:

gatherers/censys.py
```diff
@@ -1,13 +1,14 @@
-import os
-import json
 import csv
+import json
 import logging
+import os
```
konklone: Just curious, was this to alphabetize the imports?

gatherers/gatherersabc/gathererabc.py
```diff
@@ -0,0 +1,14 @@
...
+class Gatherer(metaclass=ABCMeta):
+
+    def __init__(self, suffixes: List[str], options: dict, extra: dict={}):
+        self.suffixes = suffixes
+        self.options = options
+        self.extra = extra
+
+    @abstractmethod
+    def gather(self):
+        pass
```
konklone: Is this a stray file? Looks unused, and like an older copy of `gatherers/gatherersabc.py`.

gatherers/censys.py
tadhg-ohiggins: Yes, I'm just used to having them alphabetized at this point. Alphabetized and in chunks of standard library/third-party/local.

PR #231, opened 2018-03-29
WIP: Creating a Python package

IanLee1521:
TODO:

- [x] Need to shuffle around several of the directories to make this work for real.

### DIFF
--- a/gather
+++ b/gather
@@ -55,14 +55,14 @@ def run(options=None, cache_dir="./cache", results_dir="./results"):
 
         try:
             gatherer_module = importlib.import_module(
-                "gatherers.%s" % source)
+                "domain_scan.gatherers.%s" % source)
             gatherer = gatherer_module.Gatherer(suffixes, options, extra)
         except ImportError:
             # If it's not a registered module, allow it to be "hot registered"
             # as long as the user gave us a flag with that name that can be
             # used as the --url option to the URL module.
             if options.get(source):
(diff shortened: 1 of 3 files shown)


IanLee1521:
This actually appears to be working now as is, but I'm going to use this as an excuse to roll this into a proper package structure while I'm touching it.

konklone:
@IanLee1521 This is great progress. Thank you for working on this!

Also, @jsf9k and I just maybe-simplified or maybe-complicated this work, by splitting out the requirements into different places in #224.

In a `pip install`-able use case, should it install the scanner and gatherer-specific reqs by default? This would have been the default previously, so keeping it that way by installing from all 3 `.txt` files would at least be the status quo.

I don't think we yet have a live downstream use case of the `pip install` workflow for `domain-scan`, so I guess I'd look to @jsf9k or @JasonMWhite (or others) for suggestions on what the right interface is here.

jsf9k:
I think we could keep the requirements split up by adding `gatherers`, `scanners`, etc. sections to `extras_require` in `setup.py`.  

This is what I've done with the testing requirements in `pshtt`, [for example](https://github.com/dhs-ncats/pshtt/blob/develop/setup.py#L71-L77).  `pip install .` will install the `install_requires`, and if you want to install those requirements _and_ the ones for testing then you just do `pip install .[test]`.

PR #234, opened 2018-04-02
PSL now stored in same cache location in lambda and local

jsf9k:
Now storing the PSL in the same cache location when running in Lambda and when running locally


jsf9k:
FWIW, I'm currently using this code in production.  This is ready to be merged.

PR #236, opened 2018-04-02
Adjust Dockerfile to align with recent merges

jsf9k:
In the light of recent merges, some changes to the Dockerfile are necessary.


konklone:
Thank you!

PR #237, opened 2018-04-03
Upgrade sslyze to 1.4+ versions

jsf9k:


lambda/lambda_handler.py
```diff
@@ -15,7 +15,7 @@ def handler(event, context):
 
     # Log all sent events, for the record.
     utils.configure_logging(options)
-    logging.warn(event)
+    logging.info(event)
```
konklone: :+1:

konklone:
Yay for 1.4!

PR #238, opened 2018-04-03
Some minor cleanup

jsf9k:


requirements-scanners.txt
```diff
@@ -9,7 +9,6 @@ git+https://github.com/dhs-ncats/trustymail.git#egg=trustymail
 
 # sslyze
 sslyze>=1.3.4,<1.4.0
-cryptography
```
konklone: I've liked calling out the `cryptography` dependency separately since we import it in the `sslyze` scanner independently from sslyze itself, to parse certain results. But at the same time, we'll always need the version of `cryptography` that `sslyze` is relying on to produce these structures, so it makes sense to rely on the downstream dependency. Good call. :+1:

PR #239, opened 2018-04-04
Fix some issues with gather options parsing with hyphenated sources

konklone:
Gatherer names like `censys-snapshot` were failing, because of argparse's automatic hyphen->underscore conversion. This led to `nargs=1` not being set for them, as well as the `gather` script not matching up the stated source name with the keys in the `options` dict.

This fixes that, and also ditches one of the restrictions around checking whether some flags are gatherer-specific. It seemed unnecessary to keep this mapping maintained over the long run.

### DIFF
--- a/scanners/analytics.py
+++ b/scanners/analytics.py
@@ -18,9 +18,10 @@ def init(environment, options):
     # It's a URL, download it first.
     if analytics_file.startswith("http:") or analytics_file.startswith("https:"):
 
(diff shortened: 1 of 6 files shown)


konklone:
@tadhg-ohiggins I can't figure out why the tests are failing or how to fix them. It makes sense that tests would fail, since I changed some behavior, but I'm at a loss for how to understand the test output or which cases are failing.

konklone:
Added another commit fixing `--analytics` for the `analytics` scanner.

PR #241, opened 2018-04-06
Abandon OO experiment, refactor scan code path, add tests

tadhg-ohiggins:
We should hand things off,
Not seek to carry them within.
Separate each task.

Remove OO execution paths, refactor and add tests for various parts of scan execution flow, refactor and add tests for analytics scanner, add support for explicit argument handling for each scanner as a separate function.

After struggling with making an OO-based approach work and running into too many issues, and considering that handing functions off to Lambda requires them to have all context available to them anyway, I realized that the only benefit we were getting from the OO approach was the elegance of abstract methods enforcing what methods were mandatory for scanners. Instead of that elegance, we now have a check when the scanner module is loaded to verify their presence—but less OO overhead.

I gave this the do_not_merge_yet label both because there are quite a few changes to the main code path that might break corner cases I've missed, and to leave time for arguments for returning to the OO approach.

### DIFF
--- a/scan
+++ b/scan
@@ -3,9 +3,7 @@
 import os
 import uuid
 import sys
-import glob
 import time
-import copy
 import logging
 import shutil
 import csv
@@ -14,8 +12,8 @@ import boto3
 import botocore
 from pathlib import Path
 from concurrent.futures import ThreadPoolExecutor
-from scanners.scannerabc import ScannerABC
-from typing import List
+from typing import Any, List, Tuple
+from types import ModuleType
 
 from scanners.headless.local_bridge import headless_scan
 from utils import scan_utils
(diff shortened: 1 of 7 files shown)


scan
```diff
@@ -111,118 +109,77 @@ def run(options=None, cache_dir=None, results_dir=None):
...
 
     # Run through each scanner and open a file and CSV for each.
     handles = {}
+    durations = {}
+    scan_uuid = str(uuid.uuid4())
+    # Store scan UUID.
+    logging.debug("[%s] Scan UUID." % scan_uuid)
     for scanner in scanners:
         name = scanner.__name__.split(".")[-1]  # e.g. 'pshtt'
-        scanner_filename = "%s/%s.csv" % (results_dir, name)
```
konklone: Noting that there was a regression here - this line was remove, and replaced in `scan_utils` with `scanner_csv_path = Path(results_dir, name).resolve()`, which was missing the `.csv` extension.

PR #242, opened 2018-04-09
Using PyPI instead of pulling from GitHub

jsf9k:
Now that deployment to PyPI is back on track for `pshtt` and `trustymail`, we can use explicit version numbers here instead of just pulling whatever is in `develop`.


jsf9k:
That was fast!  Thanks @IanLee1521!

PR #246, opened 2018-04-16
Fix for appending .csv to results CSV filenames

konklone:
This fixes a regression in #241 that dropped the `.csv` extension from output files in `results/` during `scan` events.


konklone:
Added a commit which fixes `handler` dict access (which is now keyed by name instead of by scanner module) for the `--lambda` code path.

konklone:
Added a commit that removes `scan` from the required attributes for scanners, since it is not required for scanners that use headless Chrome.

konklone:
@tadhg-ohiggins Mind giving this a look?

PR #247, opened 2018-04-16
Making it so that the logging level can be adjusted when running in Lambda

jsf9k:
In the case of AWS Lambda, the root logger is used BEFORE our Lambda handler runs, and this creates a default logging handler that goes to the console.  Once logging has been configured, calling `logging.basicConfig()` has no effect.  The only way to get around this is to manually remove any root handlers (if present) before calling `logging.basicConfig()`.  This unconfigures logging and allows `--debug` to affect the logging level that appears in the CloudWatch logs.

See [here](https://stackoverflow.com/questions/1943747/python-logging-before-you-run-logging-basicconfig) and [here](https://stackoverflow.com/questions/37703609/using-python-logging-with-aws-lambda) for more details.

I verified that, using the code in this pull request, I get tons more output from CloudWatch when running scanners in Lambda.  If the `--debug` flag is omitted then the output remains as it has been until now.

### DIFF
--- a/utils/utils.py
+++ b/utils/utils.py
@@ -305,6 +305,25 @@ def configure_logging(options=None):
         print("Invalid log level (specify: debug, info, warn, error).")
         sys.exit(1)
 
+    # In the case of AWS Lambda, the root logger is used BEFORE our
+    # Lambda handler runs, and this creates a default handler that
+    # goes to the console.  Once logging has been configured, calling
+    # logging.basicConfig() has no effect.  We can get around this by
+    # removing any root handlers (if present) before calling
+    # logging.basicConfig().  This unconfigures logging and allows
+    # --debug to affect the logging level that appears in the
+    # CloudWatch logs.
+    #
+    # See
+    # https://stackoverflow.com/questions/1943747/python-logging-before-you-run-logging-basicconfig
+    # and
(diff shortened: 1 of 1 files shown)


jsf9k:
I noticed this because I'm trying to debug some handshake failures in `sslyze`.

utils/utils.py
```diff
@@ -305,6 +305,25 @@ def configure_logging(options=None):
         print("Invalid log level (specify: debug, info, warn, error).")
```
jsf9k: I noticed that line 302 says `options.get("log", "warn")`.  The `--log` option does not appear to be supported by the current CLI code.  Should it be added, or should this `else` statement be removed?

tadhg-ohiggins: `--log` is still supported by the options for `./gather`, but not for `./scan`. I might have misunderstood where that's called from and am happy to add the option back to the argument handling for `./scan` if that's necessary.

jsf9k: I'll defer to @konklone, but I think that would be a useful option for `./scan` too.

konklone: I don't even remember adding this option, and can't recall ever using it. :) It's not documented [in the README](https://github.com/18F/domain-scan#options), and probably dates way back to the first alpha versions in 2014/2015.

If no one's using it, I'm comfortable removing the option entirely.

PR #255, opened 2018-05-04
Add fast caching

jsf9k:
This pull request adds support for fast (currently in-memory) caching for scanners.  The only scanner that currently supports this is sslyze.

I have found that SMTP servers (as compared to HTTP/HTTPS servers) are _much_ more sensitive to having multiple connections made to them.  In testing the various ciphers sslyze makes a lot of connections, and multiple domains often use the same SMTP servers, so it makes sense to check if we have already scanned this mail server when testing a different domain.

This is a complete redo of #254.


scan
```diff
@@ -290,11 +321,28 @@ def perform_scan(params: Tuple[Any, str, dict, dict, dict]):
...
             meta['end_time'] = scan_utils.local_now()
             meta['duration'] = meta['end_time'] - meta['start_time']
 
         if data:
+            if hasattr(scanner, 'fast_cache_support'):
+                # Update the cache
+                if FAST_CACHE_KEY not in environment:
+                    environment[FAST_CACHE_KEY] = []
+                # I hate having to use isinstance here, but I don't
+                # see another way
```
konklone: Didn't we standardize the return value as a list (of lists) here? (It can also return `None`.)

scan
```diff
@@ -88,6 +88,10 @@ LAMBDA_DETAIL_HEADERS = [
 # about Lambda executions to show up after task completion.
 LAMBDA_LOG_DELAY = 20
 
+# The environment dictionary key whose correpsonding value is the fast
```
konklone: Typo: `correpsonding` -> `corresponding`.

jsf9k:
I'll get back to this, probably next week.

jsf9k:
@konklone I know it's been forever since this PR was created, but I was revisiting it today.  I'm not understanding your proposed solution, and I had the same difficulty before.  I think I'm missing some key point.

The problem I see is that the only place in `scanners/sslyze.py` where I can modify the fast cache is inside `init_domain()`.  I can't update the fast cache when I get results back in `scan()` because [the fast cache has been removed from the environment at that point](https://github.com/18F/domain-scan/blob/improvement/add_cache_to_environment/scan#L323-L325) in order to avoid transferring it to Lambda.

Furthermore, my intent was to use this fast cache solution for `trustymail` as well since it has similar problems in thathe mail servers that are actual scanned are sometimes repeated across domains with completely different names.  I think it is useful for this feature to be (like Lambda) available for opt-in by other scanners if folks choose to support it.

Anyway, I'm hoping to reopen this pull request for discussion since I really need this feature in `trustymail` and `sslyze`. :grin:

scan
jsf9k: I don't think that's true.  Neither `scanners/pshtt.py` nor `scanners/trustymail.py`return a `list`.

konklone: Ah, I was thinking of `to_rows` here.

scanners/sslyze.py
```diff
@@ -36,15 +37,55 @@
...
+    dict
+        The data found in the fast cache corresponding to the
+        specified mail server, or None if no such data is found.
+    """
+    hostname_and_port = mail_server.split(':')
+    hostname = hostname_and_port[0]
+    port = hostname_and_port[1]
+    for record in fastcache:
+        if record['starttls_smtp'] and record['hostname'] == hostname and record['port'] == port:
+            return record
```
konklone: This part is really the thing that sticks out the most: it's iterating through a list and interrogating each value, rather than doing a dict lookup. For something like this, I would expect to see the fastcache itself be a dict with mail hosts as the dict keys. Would that be reasonable? If so, that would be simpler code, and much faster.

konklone:
> @konklone I know it's been forever since this PR was created, but I was revisiting it today. I'm not understanding your proposed solution, and I had the same difficulty before. I think I'm missing some key point.

Well, reviewing my comments and the code, I don't totally get all of my own past comments either. But the main one that still makes sense to me is the inefficiency of iterating through the cache list each time, as opposed to using a dict. Something like `environment['fastcache']['mail.example.gov']`.

scanners/sslyze.py
jsf9k: In order to do that I would have to have the fastcache `dict` available to me in the `scan()` function of `scanners/sslyze.py` and I don't.  `scan()` is called by either `perform_local_scan()` or `perform_lambda_scan()` of the file `scan`, and the fastcache `dict` isn't passed to those functions precisely because we don't want such a (potentially) large `dict` to be pushed up to AWS Lambda for every Lambda execution.

An alternative is to put scanner-specific code in `scan` in lines 333-242 where it updates the fastcache, but that's even uglier.  And it breaks the modularity of the code since `scan` should be agnostic to what is inside the individual `scanners/*.py`.

I get that it would be better and more efficient to structure the fastcache `dict` differently.  I just didn't see a decent way to do that with the way scanners currently operate in `domain-scan`.

One possible solution, which admittedly I haven't thought through completely, might be to add a post-scan lifecycle method to the scanners.  `scan` could pass in the original environment (i.e. without removing the fastcache data structure) and the results just returned by `scan` and let the scanner itself update the fastcache data structure however it sees fit.  Do you think this would address your concerns?  Maybe this is what you had in mind all along and I was just missing the point.

konklone: Got it, I now completely see the challenge in doing it the way I'm recommending. I guess the logical conclusion here is to either add a lifecycle method just for cache management (which feels heavy) or to let scanners absorb the complexity of knowing what their own data structures were as they iterate through a large cache (which is this PR).

Another option might be to allow each scanner to return a `to_cache` key that is a dict with one or more key/value pairs to toss/merge into the fastcache dict, which the `scan` orchestrator can do without knowing what they mean. This would also dramatically slim the size of the data stored in the fastcache key, saving on memory as well as speed. I imagine it wouldn't slim it to the size where it would be acceptable to pass to each Lambda function - but maybe.

Any thoughts on that? That's the last suggestion I'll throw your way, and if you don't like it, I'm happy to merge this and move on.

jsf9k: I think the lifecycle method is the cleanest way to implement it.  I could make it so that the lifecycle method is called if (1) the scanner provides it and (2) the `--in-memory-cache` flag is present when `scan` is called.

I'll implement that now and commit it to this pull request.

jsf9k: I ended up creating a CLI flag to _disable_ fast caching, since I think normally users will want it.

jsf9k:
I will fix the failing tests after lunch.  The tests that fail appear to be those that test the `scan` command line options, so I suspect they are failing because I added an option.

jsf9k:
I think this pull request is finally ready for review and merging.

scan
```diff
@@ -330,16 +327,9 @@ def perform_scan(params: Tuple[Any, str, dict, dict, dict]):
...
-                    environment[FAST_CACHE_KEY] = []
-                # I hate having to use isinstance here, but I don't
-                # see another way
-                if isinstance(data, list):
-                    environment[FAST_CACHE_KEY].extend(data)
-                else:
-                    environment[FAST_CACHE_KEY].append(data)
+            # Run the post-scan hook if it's present
+            if hasattr(scanner, 'post_scan'):
+                scanner.post_scan(domain, data, environment)
```
konklone: Should the `post_scan` method also run even if `data` is `None`? I went through a bunch of scanners and couldn't find an example of a scanner returning `None` intentionally, but I would expect that even if there was an error in the interactions between the client and Lambda, that the ability to modify the fast cache (or take other post-scan actions) might be useful.

utils/scan_utils.py
```diff
@@ -421,6 +421,11 @@ def build_scan_options_parser() -> ArgumentParser:
     parser.add_argument("--workers", nargs=1,
                         help="Limit parallel threads per-scanner to a number.")
     # TODO: Should workers have a default value?
+    parser.add_argument("--no-fast-cache", action="store_true", help="".join([
+        "Do not use fast caching even if a scanner supports it.  This option ",
+        "will cause domain-scan to use less memory, but some scans may be ",
```
konklone: Nit - it's not that some `scans` may be repeated, in the sense that a scan is a domain-scan construct of "taking an action on a hostname", but it may cause some network activity (or other expensive actions) to be needlessly repeated.

utils/scan_utils.py
```diff
@@ -440,16 +440,49 @@ def build_scan_options_parser() -> ArgumentParser:
...
+    ]))
+    parser.add_argument("--mx", help="".join([
+        "trustymail: Only check MX records"
+    ]))
+    parser.add_argument("--spf", help="".join([
+        "trustymail: Only check SPF records"
+    ]))
+    parser.add_argument("--dmarc", help="".join([
+        "trustymail: Only check DMARC records"
+    ]))
```
konklone: Thanks for adding all these docs!

scan
jsf9k: This makes a lot of sense, especially since `post_scan()` is intended to be more general than just updating an in-memory cache.  It's a lifecycle method that may be used for any purpose whatsoever.

I'm committing a change now to allow fro it.

utils/scan_utils.py
jsf9k: Change committed.

utils/scan_utils.py
jsf9k: No problem at all :smiley:

jsf9k:
@konklone, have a quick look at my most recent two commits when you have time.  If you're happy with them then I'll merge.

konklone:
Awesome, looks great! :shipit:

PR #257, opened 2018-05-24
Get a11y and third_party scans working again

micahsaul:


micahsaul:
Also made a couple small tweaks to process_a11y to get it running again.

PR #262, opened 2018-06-15
Relying on Python 3.6+ now

### LINKED ISSUE #261: Path.resolve() throws exception when results file doesn't exist
From a fresh clone, running the pshtt scanner with Python 3.5.3:

`scan domains.csv --scan=pshtt --workers=36`

```
Traceback (most recent call last):
  File "./scan", line 560, in <module>
    run(options, unknown, cache_dir, results_dir)
  File "./scan", line 120, in run
    scan_domains(scans, domains, options)
  File "./scan", line 150, in scan_domains
    scanner, options, (PREFIX_HEADERS, LOCAL_HEADERS, LAMBDA_HEADERS))
  File "cthomp/HTTP-Bad-Search-Console-Message/domain-scan/utils/scan_utils.py", line 572, in begin_csv_writing
    scanner_csv_path = Path(results_dir, "%s.csv" % name).resolve()
  File "/usr/lib/python3.5/pathlib.py", line 1109, in resolve
    s = self._flavour.resolve(self)
  File "/usr/lib/python3.5/pathlib.py", line 330, in resolve
    return _resolve(base, str(path)) or sep
  File "/usr/lib/python3.5/pathlib.py", line 315, in _resolve
    target = accessor.readlink(newpath)
  File "/usr/lib/python3.5/pathlib.py", line 422, in readlink
    return os.readlink(path)
FileNotFoundError: [Errno 2] No such file or directory: 'cthomp/HTTP-Bad-Search-Console-Message/domain-scan/results/pshtt.csv'
```
Looking at `begin_csv_writing`:
https://github.com/18F/domain-scan/blob/ad6ea2fadc79500eb2060f4f281550f32127e47a/utils/scan_utils.py#L572

The `Path.resolve()` call can't succeed since it throws `FileNotFoundError` if the file doesn't exist (https://docs.python.org/3.5/library/pathlib.html#pathlib.Path.resolve), and `scan` tries to delete the file if it already exists before this. 

Removing the call to `resolve()` allows the scan to work. (Or, if `results/` is guaranteed to exist, then this could be `Path("results/").resolve().joinpath("pshtt.csv")`.)

If results/pshtt.csv already exists, however, the following error occurs:

```
Traceback (most recent call last):
  File "./scan", line 560, in <module>
    run(options, unknown, cache_dir, results_dir)
  File "./scan", line 120, in run
    scan_domains(scans, domains, options)
  File "./scan", line 134, in scan_domains
    os.remove(result)
TypeError: remove: illegal type for path parameter
```

Which implicates:
https://github.com/18F/domain-scan/blob/ad6ea2fadc79500eb2060f4f281550f32127e47a/scan#L131-L134

Inspecting, it appears that `type(result)` is `PosixPath`, but I think `os.remove()` expects a string, so this should maybe be `os.remove(str(result))` (https://docs.python.org/3.5/library/os.html#os.remove).

konklone:
Newer versions of `domain-scan` now rely on Python 3.6+, due to the use of `Path.resolve()` that was added in some recent-ish PRs. Since upgrading from 3.5->3.6 should be pretty painless these days, it seems easier to just bump the documented minimum requirement to match.

Fixes #261.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,44 +2,6 @@
 
 version: 2
 jobs:
-  build_35:
-    docker:
-      - image: circleci/python:3.5
-
-    steps:
-    - checkout
-
-    # Download and cache dependencies
-    - restore_cache:
-        keys:
-        - v1-python35-dependencies-{{ checksum "requirements.txt" }}
-        # fallback to using the latest cache if no exact match is found
-        - v1-python35-dependencies-
-
-    - run:
-        name: install dependencies
-        command: |
-          python3 -m venv venv
-          . venv/bin/activate
-          pip install --upgrade setuptools
-          pip install flake8
-          pip install -r requirements.txt
-
-    - save_cache:
-        paths:
-          - ./venv
-        key: v1-python35-dependencies-{{ checksum "requirements.txt" }}
-
-    - run:
-        name: run tests
-        command: |
-          . venv/bin/activate
-          ./test.sh
-
-    - store_artifacts:
-        path: test-reports
-        destination: test-reports
-
   build_36:
     docker:
       - image: circleci/python:3.6
@@ -82,5 +44,4 @@ workflows:
   version: 2
   build_and_test:
     jobs:
-      - build_35
       - build_36
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ There is also built-in support for using **headless Chrome** to efficiently meas
 
 ### Requirements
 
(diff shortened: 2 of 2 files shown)


jsf9k:
You'll need to change the CircleCI configuration too, right?

konklone:
@jsf9k Good catch - I'd actually done a search for `3.5` in the repo, but it didn't turn that up. I updated the PR with that change. I also can't think of another place where it would be present.

PR #264, opened 2018-06-25
add accepted cipher list to the sslyze results

buckley-w-david:
This PR adds the list of accepted ciphers to the output of the sslyze scanner to sync up the CDS fork of domain-scan as per suggestion of @konklone


konklone:
Thanks for submitting this upstream. I'd love to see us relying on the same repo together for this layer of the pipeline.

PR #265, opened 2018-07-05
Beginnings of a USWDS scanner

konklone:
The simplest thing to start: a simple HTML string match, looking for instances of inaccurate banner text.

This is powered by headless Chrome, and could also be expanded to look for different kinds of USWDS-related use and activity.

cc @maya for ideas for what to look for.


konklone:
Works well!

PR #267, opened 2018-07-13
Start attempting to measure USWDS presence

konklone:
This is a first pass at detecting USWDS on federal websites.

@el-mapache @maya

### DIFF
--- a/scanners/uswds.js
+++ b/scanners/uswds.js
@@ -36,14 +36,27 @@ module.exports = {
       else throw exc;
     }
 
+    // Do a quick text match on the HTML to find out of date USWDS
+    // banner text.
     const html = await page.content();
-
     data.banner_bad_text = hasBadText(html);
+
+    // Search DOM for evidence of USWDS being present at all.
+    data.present = await uswdsPresent(page);
 
     return data;
   }
 }
 
+// Simple text match on some old text.
 var hasBadText = (html) => {
   return (html.search("Federal government websites always use a .gov or .mil domain.") >= 0)
+};
+
+
+// Checks for any element with a class with "usa-" in it.
+// TODO: Should "usa-" at least be at the start of the string?
+var uswdsPresent = async (page) => {
+  var match = await page.$('[class*=usa-]');
+  return (match != null);
 };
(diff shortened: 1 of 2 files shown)


scanners/uswds.js
```diff
@@ -36,14 +36,27 @@ module.exports = {
...
 }
 
+// Simple text match on some old text.
 var hasBadText = (html) => {
   return (html.search("Federal government websites always use a .gov or .mil domain.") >= 0)
 };
+
+
+// Checks for any element with a class with "usa-" in it.
+// TODO: Should "usa-" at least be at the start of the string?
```
el-mapache: I don't think so, as someone could conceivably have a class string like `my-other-semantic-class usa-banner`

maya: Agree, I wish there was a way to look for the start of a selector no matter where it's located, but don't think that exists, right?

konklone: Well, I don't know what's supported in terms of selectors, but you could do it one of two ways:

* If word boundaries are supported, then you could use something like `\b` to match on either "usa-*" at the start of the string, or after a space.
* If not, you could union two patterns together: `^usa-` and ` usa-`, and match on either of them. That would exclude `pupusa-` but still include `semantic-class usa-banner` and `usa-banner`.

Do you know which of those are doable?

el-mapache: So, css regexs are pretty bare bones compared to most other implementations. You basically have three options:
* `attr*=string` - attribute value contains the string
* `attr^=string` - attribute value begins with the string
* `attr$=string` - attribute value ends with the string

We can do something like `[class^=usa-], [class*=usa-]` which I don't think will give us the union behavior we want, e.g. it might still match `pupusa-`

konklone: You couldn't do `[class*= usa-]` (note the space) or `[class*=\susa-]` to insist that the space be there?

IanLee1521: @konklone -- Is this resolved and I can review and merge, or are you still working through the "what to do" part?

konklone: I think we can merge this in and move on. :+1:

konklone:
@IanLee1521 Want to approve and merge?

PR #269, opened 2018-08-26
Add INSTALLATION.md

jsf9k:


IanLee1521:
I know Docker is quickly taking over, but would a Vagrant / Vagrant installation be useful here as well? I'm thinking of as a supplement to the "install on Linux" part of the file. Or just another thing to keep up to date and not worth it?

jsf9k:
I'm not a big Vagrant user.  If someone else wants to maintain that part of the instructions I'm not opposed to it being in there, though.

konklone:
That looks good - please open a PR!

PR #273, opened 2018-09-13
Cast the port string into an integer

jsf9k:
sslyze 2.0 is pickier about types, and passing the port as a string here was causing sslyze to fail.

For HTTPS scanning the port is not specified (the default of 443 is used), so this bug was only impacting folks like me who are using sslyze against other ports.


jsf9k:
Once #275 gets merged, just merge in `master` and the CircleCI build should pass.

jsf9k:
Thanks @IanLee1521!

PR #276, opened 2018-09-14
Remove now-broken hack and replace with proper solution

jsf9k:
Originally [a hack was put in place to disable automatic Lambda retries](https://github.com/boto/boto3/issues/1104) because `boto3` lacked support for this.  Unfortunately this hack no longer works with the latest `boto3` code because it touches internal variables that apparently have been removed or renamed.  Fortunately, though, [support for disabling automatic retries has been added](https://github.com/boto/botocore/pull/1260).

### DIFF
--- a/scan
+++ b/scan
@@ -43,13 +43,11 @@ def set_aws_credentials(options):
 
     invoke_config = botocore.config.Config(
         max_pool_connections=global_max_workers,
-        connect_timeout=300, read_timeout=300
+        connect_timeout=300,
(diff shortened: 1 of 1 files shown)


jsf9k:
This is a pretty big deal, since `domain-scan` is broken right now if you do a fresh install or upgrade `boto3` to the latest.

IanLee1521:
See also: https://github.com/boto/botocore/pull/1260

PR #278, opened 2018-10-22
Update to use the latest pshtt, trustymail, and sslyze

jsf9k:


konklone:
I ran the tests myself to validate the build, so I'll merge so we're not waiting on GitHub to work its way through the build backlog.

PR #289, opened 2018-12-03
Add fast cache support for trustymail

jsf9k:
This pull request adds fast cache support for the `trustymail` scanner.  This is used to cache the SMTP/STARTTLS test results by mail server and port.  This is useful because in many cases multiple domains will use identical mail servers, and testing them multiple times is more likely to trigger defenses and lead to poor results.


jsf9k:
`dnspython` should be added to one of the `requirements*.txt` files, but I'm not sure which one.  It is used by the part of `scanners/trustymail.py` that is run on the host (not in AWS Lambda).

PR #292, opened 2018-12-19
Use latest trustymail

jsf9k:
The latest trustymail increases the verbosity for the SPF library that is used under the hood.  This information is occasionally useful when debugging BOD 18-01 SPF issues.


IanLee1521:
Looks good, thanks @jsf9k !

PR #295, opened 2019-02-11
remove gemnasium. 18F no longer uses gemnasium

eddietejeda:


IanLee1521:
Thanks @eddietejeda ! Good catch!

PR #296, opened 2019-02-27
Do not hard code the DNS servers to use for the trustymail scanner

jsf9k:
These should be specified on the command line if needed.

I've tested this with the NCATS BOD 18-01 scanning and it works great for our purposes.  The reason I made this change is that we were running into a connection tracking limit in AWS, and using the AWS DNS instead of reaching out externally to Google reduces the problem considerably.

### DIFF
--- a/scanners/trustymail.py
+++ b/scanners/trustymail.py
@@ -38,10 +38,10 @@ def init_domain(domain, environment, options):
             int(port)
             for port in options.get('smtp_ports', default_smtp_ports).split(',')
         }
-        dns_hostnames = options.get('dns')
-        if dns_hostnames is not None:
-            dns_hostnames = dns_hostnames.split(',')
-        resolver = dns.resolver.Resolver(configure=dns_hostnames is None)
+        dns_hostnames = list_from_dict_key(options, 'dns')
+        resolver = dns.resolver.Resolver(configure=not dns_hostnames)
+        if dns_hostnames:
+            resolver.nameservers = dns_hostnames
         # This is a setting that controls whether we retry DNS servers
         # if we receive a SERVFAIL response from them.  We set this to
         # False because, unless the reason for the SERVFAIL is truly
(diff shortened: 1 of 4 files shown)


scanners/trustymail.py
```diff
@@ -64,7 +60,8 @@ def init_domain(domain, environment, options):
         # http://www.dnspython.org/docs/1.14.0/dns.resolver-pysrc.html#Resolver._compute_timeout.
         resolver.timeout = float(timeout)
         resolver.lifetime = float(timeout)
-        resolver.nameservers = dns_hostnames
+        if dns_hostnames is not None:
```
IanLee1521: What happens if it *IS* `None` ?

Should there be a fallback / default value still?

scanners/trustymail.py
```diff
@@ -42,10 +38,10 @@ def init_domain(domain, environment, options):
             int(port)
             for port in options.get('smtp_ports', default_smtp_ports).split(',')
         }
-        dns_hostnames = options.get('dns', default_dns).split(',')
-        # Note that we _do not_ use the system configuration in
-        # /etc/resolv.conf.
-        resolver = dns.resolver.Resolver(configure=False)
+        dns_hostnames = options.get('dns')
```
IanLee1521: Why do we duplicate this `options.get(...)` and `...split(',')` when the same happens below at https://github.com/18F/domain-scan/pull/296/files#diff-37d61b45847fea9081213c0878237857R131 ?

scanners/trustymail.py
jsf9k: If it is `None` then the system DNS servers as defined in `/etc/resolv.conf` will be used.

scanners/trustymail.py
jsf9k: `init_domain()` runs on the local machine, while `scan()` potentially runs in a different Python interpreter in AWS Lambda.

I will pull the common functionality out into its own function.

jsf9k: There was actually an existing function I could use for this.  See b840473.

jsf9k:
Thanks @IanLee1521!  Now if only CircleCI would cooperate :)

IanLee1521:
Yup, I'm sitting here with it open watching!

PR #310, opened 2019-08-01
Update sslyze and pshtt versions

jsf9k:
This is a little tricky, since sslyze renamed some class member variables in version 2.1.1.


jsf9k:
These changes have now been tested and are ready to be merged.

PR #315, opened 2019-08-22
Take a new approach to scoring

waldoj:
This adds a new metric (table presence), removes a couple of unused ones, and manually assigns scores based on the presence of elements instead of tallying them.


timothy-spencer:
If you can fix the linter error that is shown in the circleci build, I would love to merge this in!

waldoj:
Ah, that's what comes of knocking off work before the build is done. :) All fixed now.

PR #318, opened 2020-04-13
Pin sslyze version to 2.x, and note Python compatibility

danielnaab:
`sslyze` is currently pinned to `> 2.1.3`, which lead to a recently released `3.0` version being installed. This version currently has an incompatibility with `domain-scan`.

Additionally, `sslyze 2.1.x` is incompatible with Python 3.8. https://github.com/nabla-c0d3/sslyze/issues/408

So this PR notes the incompatibility and pins us to `sslyze 2.x`. I'll add an issue to follow-up on making the scanner compatible with `3.x`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ There is also built-in support for using **headless Chrome** to efficiently meas
 
 ### Requirements
 
-`domain-scan` requires **Python 3.6 and up**.
+`domain-scan` requires **Python 3.6 or 3.7**.
 
 To install **core dependencies**:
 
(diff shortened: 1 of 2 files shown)


timothy-spencer:
I'm happy to approve this, but you should consider getting https://github.com/18F/domain-scan/pull/307 passing tests and merged in first.  This is the branch that has all of our new scans in it.  I'm not sure why it's failing tests, though.  I may have this change already in there too.

I would have done it, but we literally just got the :thumbsup: to manage this repo along with the Digital Dashboard people just before I left.  :-(

PR #321, opened 2020-09-09
Creates SEO scanner

tbaxter-18f:
Probably mostly for search.gov, but could be useful to other folks as well as they look into their SEO-friendliness.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,5 +13,8 @@ tldextract
 # Used locally by trustymail code even when Lambda is used
 dnspython
 
+# Used by some scanners to parse pages
+beautifulsoup4
+
 # Used in Lambda functions
 -r lambda/requirements-lambda.txt
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,8 +13,5 @@ tldextract
 # Used locally by trustymail code even when Lambda is used
 dnspython
 
-# Used by some scanners to parse pages
-beautifulsoup4
-
 # Used in Lambda functions
 -r lambda/requirements-lambda.txt
--- a/scanners/seo.py
+++ b/scanners/seo.py
@@ -125,8 +125,6 @@ def scan(domain: str, environment: dict, options: dict) -> dict:
     if len(descriptions) != len(set(descriptions)):
         results['warnings']['Duplicate descriptions found'] = True
 
-    logging.warning('DEBUG: results' + str(results))
-
     logging.warning("SEO scan for %s Complete!", domain)
 
     return results
--- a/scanners/seo.py
+++ b/scanners/seo.py
@@ -2,6 +2,7 @@ import logging
 import requests
 
 from bs4 import BeautifulSoup
+from http import HTTPStatus
 
 """
 Fairly simple scanner that makes a few checks for SEO and
@@ -26,7 +27,6 @@ Possible future scope:
 """
 
 # Set a default number of workers for a particular scan type.
-# Overridden by a --workers flag. XXX not actually overridden?
 workers = 50
 
 # This is the initial list of pages that we will be checking.
@@ -71,12 +71,12 @@ def scan(domain: str, environment: dict, options: dict) -> dict:
     # TO DO: If sitemap.xml redirects to a sitemap index,
     # we're not handling that correctly yet. We need to detect that
     # redirect, then ... what? Follow from there?
-    if sitemap_status == 200:
+    if sitemap_status == HTTPStatus.OK:
         soup = BeautifulSoup(sitemap.text, 'xml')
         urls = soup.find_all('url')
         results['sitemap']['urls found'] = len(urls)
         # and how many of those URLs appear to be PDFs
-        if len(urls) > 0:
+        if urls > 0:
             pdfcount = len([u for u in urls if '.pdf' in u.get_text()])
             results['sitemap']['PDFs found in sitemap'] = pdfcount
 
@@ -86,7 +86,7 @@ def scan(domain: str, environment: dict, options: dict) -> dict:
     # Perform the "task".
     for page in environment['pages']:
         try:
-            r = requests.get("https://" + domain + page, allow_redirects=True, timeout=4)
+            r = requests.get("https://" + domain + page, timeout=4)
             htmlsoup = BeautifulSoup(r.text, 'lxml')
             # get title and put in dupe-checking list
             title = htmlsoup.find('title').get_text()
--- a/scanners/seo.py
+++ b/scanners/seo.py
@@ -1,0 +1,128 @@
+import logging
+import requests
+
+from bs4 import BeautifulSoup
+
+"""
+Fairly simple scanner that makes a few checks for SEO and 
+search indexing readiness. It will attempt to determine:
+* Does a robots.txt file exist at the expected location (/robots.txt)
+* If a sitemap.xml file exists
+  * If so, how many URLs are in it?
+  * How many of those URLs are PDF files?
+* Does a `main` element exist on pages
+* Does OG data exist (particularly date)
+
+Further, by comparing the site root with a second page (/privacy in most cases)
+it will compare the two pages to determine if the title and descriptions 
+appear to be unique.
+
+Possible future scope:
+* crawl delay (how do we determine)
+* total urls (set of nav + sitemap maybe?)
+* builtwith (requires paid API access)
+
+"""
+
+# Set a default number of workers for a particular scan type.
+# Overridden by a --workers flag. XXX not actually overridden?
+workers = 50
+
+
+# This is the list of pages that we will be checking.
+pages = [
+    "/",
+    "/privacy",
+]
+
+
+# Optional one-time initialization for all scans.
+# If defined, any data returned will be passed to every scan instance and used
+# to update the environment dict for that instance
+# Will halt scan execution if it returns False or raises an exception.
+#
+# Run locally.
+def init(environment: dict, options: dict) -> dict:
+    logging.debug("Init function.")
+    return {'pages': pages}
+
+
+# Required scan function. This is the meat of the scanner, where things
+# that use the network or are otherwise expensive would go.
+# Runs locally or in the cloud (Lambda).
+def scan(domain: str, environment: dict, options: dict) -> dict:
+    logging.debug("Scan function called with options: %s" % options)
+
+    # Initialize results with a quick check for the presence of robots.txt. 
+    # Because we're not reading robots.txt, we only need the status code.
+    # We could put these in `pages` but since we have specialized needs for each 
+    # and no need to iterate through them, I don't want to.
+    results = {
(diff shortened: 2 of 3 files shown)


scanners/seo.py
```diff
@@ -0,0 +1,146 @@
...
+# Runs locally or in the cloud (Lambda).
+def scan(domain: str, environment: dict, options: dict) -> dict:
+    logging.debug("Scan function called with options: %s" % options)
+
+    # Initialize results with a quick check for the presence of robots.txt.
+    # Because we're not reading robots.txt, we only need the status code.
+    # We could put these in `pages` but since we have specialized needs for each
+    # and no need to iterate through them, I don't want to.
+    results = {
+        'robots': requests.get("https://" + domain + '/robots.txt').status_code,
```
alexbielen, line 61: Thoughts on using `urllib.parse.urljoin` or `path.join` to build the urls? That way we can avoid having to keep track of `/`s.

scanners/seo.py
```diff
@@ -0,0 +1,146 @@
...
+    }
+    # Now let's take a look at sitemap
+    sitemap = requests.get("https://" + domain + '/sitemap.xml')
+    sitemap_status = sitemap.status_code
+    results['sitemap'] = {
+        'status': sitemap_status
+    }
+    # If we have found a sitemap, see how many URLs are in there?
+    # TO DO: If sitemap.xml redirects to a sitemap index,
+    # we're not handling that correctly yet. We need to detect that
```
alexbielen, line 72: Can you say some more about this comment?

It looks like `requests` handles redirects for `get` requests [by default](https://2.python-requests.org/en/master/user/quickstart/#redirection-and-history). But I'm not sure if the comment is getting at something else that I'm missing.

scanners/seo.py
```diff
@@ -0,0 +1,146 @@
...
+    # redirect, then ... what? Follow from there?
+    if sitemap_status == 200:
```
alexbielen, line 74: My totally pedantic take on this is....

```
from http import HTTPStatus

if sitemap_status == HTTPStatus.OK:
...
```

Then we can avoid the dreaded magic numbers.

scanners/seo.py
```diff
@@ -0,0 +1,146 @@
...
+        soup = BeautifulSoup(sitemap.text, 'xml')
+        urls = soup.find_all('url')
+        results['sitemap']['urls found'] = len(urls)
+        # and how many of those URLs appear to be PDFs
+        if len(urls) > 0:
```
alexbielen, line 79: We can simplify/pythonify this to `if urls:` and avoid a function call.

scanners/seo.py
```diff
@@ -0,0 +1,146 @@
...
+            pdfcount = len([u for u in urls if '.pdf' in u.get_text()])
+            results['sitemap']['PDFs found in sitemap'] = pdfcount
+
+    # We'll write to these empty lists for simple dupe checking later
+    titles = []
+    descriptions = []
+    # Perform the "task".
+    for page in environment['pages']:
+        try:
+            r = requests.get("https://" + domain + page, allow_redirects=True, timeout=4)
```
alexbielen, line 89: `allow_redirects` is `True` by default so it's okay to leave that out here.

scanners/seo.py
```diff
@@ -0,0 +1,146 @@
...
+
+            results[page] = {
+                'page': page,
+                'status': str(r.status_code),
+                'title': title,
+                'description': description,
+                'has main tag': maintag,
+                'DC Date': dc_date
+            }
+        except Exception as error:
```
alexbielen, line 119: I'm not totally sure, but it looks like this `except` block is handling network errors from `requests`? Thoughts on moving it up closer to the `requests.get` call? Then we can handle the different network errors as necessary.

```
try:
    r = requests.get(url, params={'s': thing})
except requests.exceptions.Timeout:
    # Maybe set up for a retry, or continue in a retry loop
except requests.exceptions.TooManyRedirects:
    # Tell the user their URL was bad and try a different one
except requests.exceptions.RequestException as e:
    # catastrophic error. bail.
    raise SystemExit(e)
```

^ From [this SO comment](https://stackoverflow.com/a/16511493).

scanners/seo.py
```diff
@@ -0,0 +1,146 @@
...
+            results[page] = "Could not get data from %s%s: %s" % (domain, page, error)
+
+    # now check for dupes
+    if len(titles) != len(set(titles)):
```
alexbielen, line 123: Thoughts on using `collections.Counter` for the dupe-checking instead? We could avoid creating the lists and the set and just create the `Counter`.

requirements.txt
```diff
@@ -13,5 +13,8 @@ tldextract
 # Used locally by trustymail code even when Lambda is used
 dnspython
 
+# Used by some scanners to parse pages
+beautifulsoup4
```
alexbielen, line 17: I believe that the dependencies for scanners are in the `requirements-scanners.txt` file, so I believe we should add it there -- but TBH I'm not sure why the requirements are architected this way. 

Also, any reason not to pin the version?

alexbielen (changes requested):
Nice, looks like it does the job! I left some comments, mostly about the `requests` library usage and some Python style things. 

One general thought is that the `scan` function is pretty long -- I'm not sure if there are any logical places to break it up though! Lemme know what you think.

requirements.txt
alexbielen: Update: looks like it already is in requirements-scanners.txt

scanners/seo.py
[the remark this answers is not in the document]
tbaxter-18f: I thought about that but I think there's a high likelihood one of the first requests will be to read robots, and figured the different type of request could be a gotcha for whoever tries to make that happen. I agree, it'd be slightly more performant but I figured consistency might be better here.

tbaxter-18f: 🤷 No real opinion either way. I was just trying to be consistent with how they're constructed elsewhere (at least in the example I started from). If you feel strongly about it, I'm happy to change it. I do agree, the presence or absence of `/` is pretty easy to get wrong.

scanners/seo.py
tbaxter-18f: Sure. Some sites have a lot of URLs, so they take advantage of the ability to break them up into multiple sitemap files. So their actual sitemap.xml becomes an index file that just lists the real sitemaps. In that case we would need to (a) detect that's what's going on and (b) follow the resulting sitemap files to get the URLs. We're not doing that yet.

scanners/seo.py
tbaxter-18f: No, I like that too. There is a potential for OK-ish status codes (like 206) that this should handle better, too.

scanners/seo.py
tbaxter-18f: Hmmm. I had it in my head that BS would return an empty ResultSet and still argue that URLs existed (albeit empty). But I think maybe I'm wrong.

scanners/seo.py
tbaxter-18f: Cruft from the copied-from scan. Might be something to watch for.

scanners/seo.py
tbaxter-18f: It's really capturing any error in getting the various data. That's why it's writing the error to results. So you can hopefully figure out what happened. In testing, most of the errors were because the page wasn't as expected -- which should be expected.

scanners/seo.py
tbaxter-18f: No thoughts. Never used it. It _feels_ like overkill given the small size of the lists, but I could easily be persuaded otherwise.

scanners/seo.py
```diff
@@ -0,0 +1,146 @@
...
+        results['warnings']['Duplicate titles found'] = True
+    if len(descriptions) != len(set(descriptions)):
+        results['warnings']['Duplicate descriptions found'] = True
+
+    logging.warning('DEBUG: results' + str(results))
```
tbaxter-18f, line 128: need to remove my testing line here.

utils/utils.py
```diff
@@ -707,5 +707,5 @@ def suffix_pattern(suffixes):
     return re.compile("(?:%s)$" % center)
 
 
-def flatten(l):
-    return list(chain.from_iterable(l))
+def flatten(iterable):
```
tbaxter-18f, line 710: This is actually outside the scope of my scanner, but that `l` var make circle unhappy.

scanners/seo.py
alexbielen: Gotcha! Thanks for clarifying that.

utils/utils.py
alexbielen: Thanks for fixing this one! Just showed up in my linter too.

scanners/seo.py
alexbielen: Sounds good. I think in the spirit of consistency we keep it as is.

scanners/seo.py
alexbielen: I don't feel strongly about this one. I think we'd need two Counters anyway to distinguish `titles` vs `descriptions so I don't think it'd actually clean it up.

tbaxter-18f: I also thought about just throwing the error the first time x is in list. Maybe. Again, no strong opinions.

scanners/seo.py
alexbielen: Gotcha. And that error data would be visible in the CSV/JSON?

scanners/seo.py
```diff
@@ -0,0 +1,144 @@
...
+    # If we have found a sitemap, see how many URLs are in there?
+    # TO DO: If sitemap.xml redirects to a sitemap index,
+    # we're not handling that correctly yet. We need to detect that
+    # redirect, then ... what? Follow from there?
+    if sitemap_status == HTTPStatus.OK:
+        soup = BeautifulSoup(sitemap.text, 'xml')
+        urls = soup.find_all('url')
+        results['sitemap']['urls found'] = len(urls)
+        # and how many of those URLs appear to be PDFs
+        if urls > 0:
```
alexbielen, line 79: If I'm reading this right, I think you'll get a `TypeError` here: 
```python
>>> urls > 0
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'ResultSet' and 'int'
```

alexbielen (changes requested):

scanners/seo.py
tbaxter-18f: You're reading it exactly right and I should know better than to rush CR fixes.

scanners/seo.py
tbaxter-18f: Should be, yeah. It's in results, which is cast to csv.

tbaxter-18f: Hold the phone. Only part of results is NOT getting to CSV. Stand by while I figure out why the rest is getting thrown out.

